PR #2, opened 2019-09-20
Initial sdk

### LINKED ISSUE #1: Extract code from hello-swap
[hello-swap](https://github.com/comit-network/hello-swap) now uses cnd's HTTP API in such a way that it would make sense to extract it into an SDK.

Tasks:
- Define interface for all swap related calls within hello-world
- Extract implementation code from hello-world into the library
- publish the library on npm
- use the library in hello-world and replace communication against cnd by using the methods provided by the library interface

DoD:
- hello-swap doesn't interface directly with cnd.

thomaseizinger:
This PR:
- copies `Cnd`, `BitcoinWallet`, `EthereumWallet` and all companion functions from `hello-world`
- Creates declaration files for all the ts code and checks in the JavaScript code

The last step is not ideal but allows people to also depend on the repository and not just on published versions of the package. For now, we need to make sure we always build before we commit but that can be optimized later :)

Resolves #1.


bonomat:
Do you want to add a travis or circle ci file right form the beginning?

PR #3, opened 2019-09-20
Add function to get cnd peer addresses

luckysori:
Otherwise you have to hardcode a peer address when building a COMIT app.

### DIFF
--- a/src/cnd.ts
+++ b/src/cnd.ts
@@ -203,7 +203,7 @@ export class Cnd {
     return info.id;
   }
 
-  public async getPeerAddresses(): Promise<string[]> {
+  public async getPeerListenAddresses(): Promise<string[]> {
     const info = await this.getInfo();
     if (!info.listen_addresses) {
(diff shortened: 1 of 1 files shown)


src/cnd.ts
```diff
@@ -203,6 +203,14 @@ export class Cnd {
     return info.id;
   }
 
+  public async getPeerAddresses(): Promise<string[]> {
```
da-kami: 🔧 Not sure this is very useful like this. 
I assume what you want to achieve in the end to be able to create the connection information for the comit-link, thus you need: 
1. The peer ID
2. Your external IP-address in the specific format

Given, that multiple IP addresses are returned in `listen_addresses` we would probably have to filter them.

Conclusion:
I would rename the function to `getIpAddressDetails` or similar.

luckysori: Good point, the name is bad.

I think I would just rename it to something like `getPeerListenAddresses` for now and delegate the choice of which one to use to the application.

PR #17, opened 2019-10-08
Remove getSwap method

thomaseizinger:
This method assumes a certain path structure on cnd side that we don't want to expose publicly.

### DIFF
--- a/src/cnd.ts
+++ b/src/cnd.ts
@@ -244,16 +244,6 @@ export class Cnd {
     });
   }
 
-  public async getSwap(id: string): Promise<EmbeddedRepresentationSubEntity> {
-    const response = await axios.get(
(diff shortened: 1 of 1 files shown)


D4nte:
> This method assumes a certain path structure on cnd side that we don't want to expose publicly.

Does it?

D4nte:
> This method assumes a certain path structure on cnd side that we don't want to expose publicly.

Does it?

thomaseizinger:
> > This method assumes a certain path structure on cnd side that we don't want to expose publicly.
> 
> Does it?

Yes, it exposes how to fetch a single swap. We should instead encourage users to either use the `Location` header for swaps they created or follow the `self` links from the response of `GET /swaps`.

PR #26, opened 2019-10-28
Minor modifications and fixes

luckysori:
Minor modifications to enable comit-network/comit-rs#1584.

Not sure if any of these merit being mentioned in the changelog. Let me know if you think it does.


src/ethereumWallet.ts
```diff
@@ -6,9 +6,11 @@ import { HDNode } from "ethers/utils/hdnode";
 export class EthereumWallet {
   private readonly wallet: Wallet;
 
-  public constructor(key: SigningKey | HDNode | Arrayish, jsonRpcUrl: string) {
+  public constructor(jsonRpcUrl: string, key?: SigningKey | HDNode | Arrayish) {
     const provider = new ethers.providers.JsonRpcProvider(jsonRpcUrl);
-    this.wallet = new ethers.Wallet(key).connect(provider);
+    const wallet = !key ? ethers.Wallet.createRandom() : new ethers.Wallet(key);
```
D4nte:
```suggestion
    const wallet = key ? ethers.Wallet.createRandom() : new ethers.Wallet(key);
```
It is a strange use of the `!` to me here, can you explain?

luckysori: If the key is undefined because it wasn't passed as an argument, create a random wallet. Otherwise, create a wallet with the key passed as an argument.

D4nte: I see, I think the usage of `!` with a ternary operator just makes it less readable.
```suggestion
    const wallet = key ? new ethers.Wallet(key) : ethers.Wallet.createRandom();
```

luckysori: Fair point!

PR #32, opened 2019-10-31
Automated release process

### LINKED ISSUE #7: Problem: Build artifacts (JavaScript files) are checked into version control
This is sub-optimal because one might forget to generate them before committing.

Two options:
- only generate the JavaScript files before releasing and bundle them for the release
- only distribute the typescript files in the npm package

Recommendation: can't give any because I don't know the state of the art approach for distributing typescript modules on npm.

### LINKED ISSUE #8: Have an automated release process
**Goal:** Have a proper release process

**Problem:** Currently, releases are done manually of master

**Recommendation:** Have an automated process that releases a new version of the package for tagged commmits.

Note:
Unless #7, is done before this issue, we need to make sure we rebuild before we release.

luckysori:
Steps to release (correct me if I'm wrong @D4nte):
1. Create branch with name matching this glob pattern: `release/*`
2. Bump version and commit changes.
3. Tag the commit with the new version number.
4. Push to remote (including tags).

To just check if the release builds and the test pass without errors, you can push to a `release/*` branch **without** a tag, [as suggested by @D4nte](https://github.com/comit-network/comit-js-sdk/pull/32#discussion_r341411430). Example outcome [here](https://github.com/comit-network/comit-js-sdk/commit/79999f8590ebc71fd5c723c292200ac88431b642/checks?check_suite_id=291117212). 

Since dealing with #8 meant having to ensure that artifacts were built before publishing I decided to also tackle #7. In my opinion, committing something that can be derived from the source files is wrong, so I added `dist/` to `.gitignore` and ran `git rm -r --cached dist/`. I also had a look at a few of the top TypeScript libraries and they also seem to bundle `.js` files for release, so I went for building and bundling during release.

Successful test run [here](https://github.com/comit-network/comit-js-sdk/commit/5a784645e8a6ad0872dc8b28ec0fdeb6e2e9c004/checks?check_suite_id=289157278).

Fixes #8.
Fixes #7.


D4nte:
> Successful test run [here](https://github.com/comit-network/comit-js-sdk/commit/5a784645e8a6ad0872dc8b28ec0fdeb6e2e9c004/checks?check_suite_id=289157278).

How come 0.4.2 is actually not published on npmjs.com?

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  push:
+    branches:
+      - 'release/*'
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
+
+jobs:
+  release:
+    if: startsWith(github.ref, 'refs/tags/')
```
D4nte: 📣 I would actually suggest to put the condition only on the release tasks.
So that when preparing a release branch, it is confirmed that it builds successfully. I would also run the tests for a final check.

If you strongly disagree then I'd just remove this `if` and the `branch` filter so that the task is only triggered on the tag as currently, it's strange to use `push` argument `branches` to only do this task on the branch but then filter all the steps out with this `if` condition.

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '10.x'
+          registry-url: 'https://registry.npmjs.org'
+      - name: Build artifacts
+        run: |
+          yarn install
+          yarn tsc
```
D4nte: 📣 Looks like `rebuild` is never called neither is `gen`, aren't they needed? Especially that you removed the `dist` folder from the repo (from what you said in the PR description) then I am guessing you need to be sure you do them.

.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules/
+dist/
```
D4nte: 📣 I think you need to do a `git rm -rf dist` as my understanding is that currently the dist folder is still checked in. Here you just stop us to update it...

.github/workflows/release.yml
luckysori: I agree with your first suggestion ☺️

luckysori:
> > Successful test run [here](https://github.com/comit-network/comit-js-sdk/commit/5a784645e8a6ad0872dc8b28ec0fdeb6e2e9c004/checks?check_suite_id=289157278).
> 
> megaquestion How come 0.4.2 is actually not published on npmjs.com? (it's a blocking question because it means to me something went wrong).

It was only a test release so I deleted it immediately from npm (and GitHub releases).

.gitignore
luckysori: You're absolutely right. I thought gitignoring would be sufficient. Thank you.

.github/workflows/release.yml
luckysori: From the `package.json`

```json
"rebuild": "rimraf dist/ && tsc"
```

This means that `yarn tsc` actually generates the artifacts. We also don't need to remove the folder since it won't exist in the repository anymore. Given that `yarn tsc` is sufficient.

D4nte: oh my god I just understood what the hell `rimraf` is. I thought it was some magic tool that generated files!

luckysori: Haha, yep I was also so confused in the beginning.

PR #35, opened 2019-11-01
Fix check

D4nte:
`yarn check` != `yarn run check`
Also, we were not doing a check in release process.
Finally, best practice to not check-in generated files.

PS: What is `postintall` https://docs.npmjs.com/misc/scripts


.github/workflows/nodejs.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: yarn install, build, and test
         run: |
           yarn
-          yarn check
+          yarn run check
```
luckysori: :mega: I believe you need to run `yarn postinstall` beforehand for this to pass. Without the files inside `gen/` it won't compile.

package.json
```diff
@@ -7,12 +7,10 @@
   "repository": "git@github.com:comit-network/comit-js-sdk.git",
   "author": "CoBloX developers <team@coblox.tech>",
   "scripts": {
-    "check": "tsc --noEmit",
+    "check": "tsc --noEmit && tslint -p . && prettier --check '**/*.{ts,js,json,yml}'",
     "rebuild": "rimraf dist/ && tsc",
-    "gen": "mkdir -p ./gen && json2ts -i ./siren.schema.json -o ./gen/siren.d.ts",
+    "postinstall": "mkdir -p ./gen && json2ts -i ./siren.schema.json -o ./gen/siren.d.ts && prettier --write ./gen/siren.d.ts",
```
luckysori: :question: Why did you call it `postinstall`?

.github/workflows/nodejs.yml
D4nte: `yarn` does `yarn install`.
The `postinstall` script is always ran automatically after `install`.

package.json
D4nte: so that it's run automatically after `install`.

.github/workflows/nodejs.yml
luckysori: Fair enough, thanks for the explanation! 

I do have to say though that I hate that it's implicit.

D4nte: well, I think it's a npm standard here to allow you to do exactly that.

PR #58, opened 2019-11-14
Remove irrelevant documentation

thomaseizinger:
The SDK can be used in a variety of ways, with examples/projects using create-comit-app being only one of them. In order to not confuse developers in what they can do with the SDK, we should not mention implementation details of how create-comit-app works here.


D4nte:
Do not merge manually! Asking mergify support for help.

PR #60, opened 2019-11-18
Add negotiation protocol and http implementation

D4nte:
Helps with comit-network/create-comit-app#128
Used in comit-network/create-comit-app#208

### DIFF
--- a/src/negotiation/execution_params.spec.ts
+++ b/src/negotiation/execution_params.spec.ts
@@ -1,0 +1,75 @@
+import moment = require("moment");
+import { ExecutionParams, validateExecutionParams } from "./execution_params";
+
+const defaultExecutionParams = () => {
+  return {
+    peer: {
+      peer_id: "QmWsfYSJ1fzeCWG7N3y2RPQMrWNXtqWPWqKqqKbXQq2MMs",
+      address_hint: "/ip4/a.b.c.d/tcp/9939"
+    },
+    alpha_expiry: moment().unix() + 48 * 60 * 60,
+    beta_expiry: moment().unix() + 24 * 60 * 60,
+    ledgers: {
+      bitcoin: { network: "mainnet" },
+      ethereum: { chain_id: 1 }
+    }
+  };
+};
+
+describe("Taker Negotiator", () => {
+  it("Validates mainnet execution parameters", () => {
+    const executionParams: ExecutionParams = { ...defaultExecutionParams() };
+
+    expect(validateExecutionParams(executionParams)).toBeTruthy();
+  });
+
+  it("Validates test execution parameters", () => {
+    const executionParams: ExecutionParams = {
+      ...defaultExecutionParams(),
+      alpha_expiry: moment().unix() + 2 * 60 * 60,
+      beta_expiry: moment().unix() + 1 * 60 * 60,
+      ledgers: {
+        bitcoin: { network: "regtest" },
+        ethereum: { chain_id: 3 }
+      }
+    };
+
+    expect(validateExecutionParams(executionParams)).toBeTruthy();
+  });
+
+  it("Invalidates mainnet execution parameters due to expiry", () => {
+    const executionParams: ExecutionParams = {
+      ...defaultExecutionParams(),
+      alpha_expiry: moment().unix() + 2 * 60 * 60,
+      beta_expiry: moment().unix() + 1 * 60 * 60
+    };
+
+    expect(validateExecutionParams(executionParams)).toBeFalsy();
+  });
+
+  it("Invalidates test execution parameters due to expiry", () => {
+    const executionParams: ExecutionParams = {
+      ...defaultExecutionParams(),
+      alpha_expiry: moment().unix() + 1 * 60 * 60,
+      beta_expiry: moment().unix() + 3 * 60 * 60,
+      ledgers: {
+        bitcoin: { network: "regtest" },
+        ethereum: { chain_id: 3 }
+      }
+    };
+
+    expect(validateExecutionParams(executionParams)).toBeFalsy();
+  });
+
+  it("Invalidates mixed network execution parameters", () => {
+    const executionParams: ExecutionParams = {
+      ...defaultExecutionParams(),
+      ledgers: {
+        bitcoin: { network: "mainnet" },
+        ethereum: { chain_id: 3 }
+      }
+    };
+
+    expect(validateExecutionParams(executionParams)).toBeFalsy();
+  });
+});
--- a/src/negotiation/execution_params.ts
+++ b/src/negotiation/execution_params.ts
@@ -1,0 +1,72 @@
+import moment = require("moment");
+
+// Information that cannot be deduced from the Order
+// and is not up for negotiation
+// Matches exactly cnd http api
+export interface ExecutionParams {
+  peer: {
+    peer_id: string;
+    address_hint: string;
+  };
+  alpha_expiry: number;
+  beta_expiry: number;
+  ledgers: LedgerParams;
+}
+
+interface LedgerParams {
+  [name: string]: {
+    network?: string;
+    chain_id?: number;
+  };
+}
+
+export function validateExecutionParams(
+  executionParams: ExecutionParams
+): boolean {
+  const now = moment().unix();
+  const relativeAlphaExpiry = executionParams.alpha_expiry - now;
+  const relativeBetaExpiry = executionParams.beta_expiry - now;
+  switch (getNetworkType(executionParams.ledgers)) {
+    case NetworkType.AllMainnet: {
+      // For mainnet, we expect 24 hours and 48 hours expiry.
+      // Checking 23 hours/47 hours cause we're lazy and it's still good enough
+      return (
+        relativeAlphaExpiry > 47 * 60 * 60 && relativeBetaExpiry > 23 * 60 * 60
+      );
+    }
+    case NetworkType.AllTest: {
+      // For testnet/regtest: don't really care as long as it makes sense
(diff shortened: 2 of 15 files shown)


src/negotiation/http/maker_client.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(makerUrl: string) {
+    this.makerUrl = makerUrl;
+  }
+
+  public async getOrder(tradingPair: string) {
+    const response = await axios.get(`${this.makerUrl}/${tradingPair}`);
+    return response.data;
+  }
+
+  public async getExecutionParams(order: Order) {
```
D4nte: This was carrying a terminology of `accept` in the example.

However, as this endpoint is readonly (`HTTP GET`) then it was confusing to have a `write` terminology of `accepting` an order.

If we were to actually tell the maker that the taker `accepts` the order, then I believe it should make a difference for the `maker`.

As we were not, I just changed that to `getExecutionParams` which is a `readonly` action (`get`).

Now, if the intent is for the maker to have an impact on the fact that a taker takes an offer, for example, by delisting the order as it has been accepted, then comment here and let me know.

In this case, I would change the terminology and actually make the maker delist its order (or marked it as accepted).

src/negotiation/maker.ts
```diff
@@ -0,0 +1,28 @@
+import { ExecutionParams, Order } from "./payload";
+
+export class Maker {
```
bonomat: This naming feels weird to me.
Right now it indicates to me that some logic is behind the implementation (a market maker comes up with a rate, etc.). 
I think you tried to separate out some concerns but I'm not sure you made it better.

src/negotiation/http/maker.ts
```diff
@@ -0,0 +1,47 @@
...
+        res.send(this.maker.getExecutionParams());
+      }
+    });
+
+    app.listen(port, () =>
+      console.log(`Maker's Negotiation Service is listening on port ${port}.`)
+    );
+  }
+
+  public addOrder(order: Order) {
```
bonomat: 📣
If you have access to the `maker` from the outside who holds all the order then you don't need this function.

D4nte:
> What state is this PR draft in?

Just pending example integration and getting some comments :)

> I have to be hones: I'm a bit confused with the naming of the classes and interfaces:
> 
>     * `Maker`: holds `orders` and `execution params` and has a function to add an additional `order`. When I hear `maker` I'm thinking of a `maker` who comes up with rates/orders/ etc.

I see your point. This class `Maker` is inside the `negotiation` module has it enables an actors to handle the `Maker` side of negotiation. The previous name was `NegotiationProtocolHandler`. However, as the protocol is asymmetric a `Handler` prefix is confusing. Does `Handler` mean the taker or the maker side?
This is why I asked at the office what terminology we used for negotiation and whether maker/taker is an appropriate terminology.

I did hesitate whether to call it `negotiation.Maker` (i.e., `Maker` inside `negotiation` module` or something like `MakerNegotiationHandler`. I'll opt for something closer to the later.

>     * `HttpMaker`: provides a rest interface to the `maker`
>     * `HttpMakerClient`: an http client to access the `HttpMaker`. How about calling this class `Taker`?

As you can see the class `Taker` execute the taker side of the protocol.

> 
>     * `MakerClient`: which lives in a file called `taker.ts` ?
> 
>     * `Taker`: why a separate class?
> 
> 
> Imho one important thing is missing (that's the main reason for blocking this):
> 
> We want the behavior to be as fluent as possible, i.e. with our current CND API: once the `taker` **took** an offer, a `swap request` is sent to the `maker` who then automatically **accepts** it.

Thank you for clarifying this, the expectations on the taker behaviour were not specified in comit-network/create-comit-app#128 :)

```

```
> This means, that the negotiation protocol should be integrated into the sdk and not just being a side product.
> 
> We should integrate this into the sdk as otherwise the examples (or other applications) would need to implement the logic which puts a lot of complexity into the application logic and onto the developers.

src/negotiation/http/maker_client.ts
D4nte: Yep, thank you for clarifying.

bonomat:
> > I have to be hones: I'm a bit confused with the naming of the classes and interfaces:
> > ```
> > * `Maker`: holds `orders` and `execution params` and has a function to add an additional `order`. When I hear `maker` I'm thinking of a `maker` who comes up with rates/orders/ etc.
> > ```
> 
> I see your point. This class `Maker` is inside the `negotiation` module has it enables an actors to handle the `Maker` side of negotiation. The previous name was `NegotiationProtocolHandler`. However, as the protocol is asymmetric a `Handler` prefix is confusing. Does `Handler` mean the taker or the maker side?
> This is why I asked at the office what terminology we used for negotiation and whether maker/taker is an appropriate terminology.
> 
> I did hesitate whether to call it `negotiation.Maker` (i.e., `Maker` inside `negotiation` module`or something like`MakerNegotiationHandler`. I'll opt for something closer to the later.

For me personally this still sounds confusing but I'm happy to roll with your intuition. 
One thing I'd change though is the `client`-side of things. Imho we can just call this `Taker` instead of `MakerClient`. 

> > 
> > Imho one important thing is missing (that's the main reason for blocking this):
> > We want the behavior to be as fluent as possible, i.e. with our current CND API: once the `taker` **took** an offer, a `swap request` is sent to the `maker` who then automatically **accepts** it.
> 
> Thank you for clarifying this, the expectations on the taker behaviour were not specified in [comit-network/create-comit-app#128](https://github.com/comit-network/create-comit-app/issues/128) :)

Sorry for the confusion, you're right it was only mentioned in the `Recommendation` :)

D4nte:
> > Thank you for clarifying this, the expectations on the taker behaviour were not specified in [comit-network/create-comit-app#128](https://github.com/comit-network/create-comit-app/issues/128) :)
> 
> Sorry for the confusion, you're right it was only mentioned in the `Recommendation` :)

The recommendation is:

> While doing this, also move the auto-accept into the comit-js-sdk

That does not say anything about the taker side of things.

I am not here to bicker but clearly you had more in your mind than there was in this issue, making the whole experience quite discouraging.

src/negotiation/http/maker.ts
```diff
@@ -0,0 +1,47 @@
+/// An Http implementation of the maker negotiation service
+
+import express from "express";
+import { Maker } from "../maker";
+import { Order } from "../payload";
+
+export class HttpMaker {
```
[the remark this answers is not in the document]
D4nte: Please check latest changes, naming has been improved.

src/negotiation/maker.ts
D4nte: Terminologies have changed, please check.

D4nte:
Reminder: Squash commits

src/negotiation/http/maker.ts
D4nte: Not sure I understood this comment so please check again and re-iterate if still not satisfied.

src/negotiation/execution_params.spec.ts
```diff
@@ -0,0 +1,89 @@
+import moment = require("moment");
+import { ExecutionParams, validateExecutionParams } from "./execution_params";
+
+describe("Taker Negotiator", () => {
+  it("Validates mainnet execution parameters", () => {
+    const executionParams: ExecutionParams = {
```
thomaseizinger: Could extract an instance with `defaultExecutionParams` (just local to this file) and then use that with spread notation to override the what is relevant to this test:

```js
const executionParams: {
   ...defaultExecutionParams(),
   // override here
};
```

src/tokens/tokens.ts
```diff
@@ -0,0 +1,16 @@
+import definitions from "./definitions.json";
```
thomaseizinger: Do you know how this is going to work once it is compiled to JS? Is this embedded into the file?
I've made bad experiences with importing json like this, especially once you try to pack everything up with webpack (important for shipping the SDK as part of a web application).

I am not sure if this particular case would work, that is why I would avoid it in general. If the list is static, just inline it into the typescript and nothing can go wrong :)
Since it is within the SDK, the user can't modify it anyway!

src/negotiation/execution_params.spec.ts
D4nte: Dang, if this syntax works then I need to use more often!

src/tokens/tokens.ts
D4nte: I wish I understood how MEW does it :)

src/negotiation/execution_params.ts
```diff
@@ -0,0 +1,72 @@
...
+// and is not up for negotiation
+// Matches exactly cnd http api
+export interface ExecutionParams {
+  peer: {
+    peer_id: string;
+    address_hint: string;
+  };
+  alpha_expiry: number;
+  beta_expiry: number;
+  ledgers: LedgerParams;
```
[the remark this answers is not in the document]
D4nte: Let's have a follow-up PR to default to mainnet, ok?

src/negotiation/maker_negotiator.ts
```diff
@@ -0,0 +1,114 @@
...
+  private ordersByTradingPair: { [tradingPair: string]: Order } = {};
+  private ordersById: { [orderId: string]: Order } = {};
+  private readonly executionParams: ExecutionParams;
+  private readonly comitClient: ComitClient;
+  private readonly tryParams: TryParams;
+
+  constructor(
+    comitClient: ComitClient,
+    executionParams: ExecutionParams,
+    tryParams: TryParams
```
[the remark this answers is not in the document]
D4nte: You need `TryParams` to accept the order.
May I suggest you try to remove them in a follow-up PR and see how it goes?

PR #64, opened 2019-11-25
Make expiries optional

### LINKED ISSUE #52: Allow swap requests without specified expiries
## Problem

The SDK enforces that a swap request must include explicit values for the expiries:

```ts
export interface SwapRequest {
  alpha_ledger: Ledger;
  alpha_asset: Asset;
  beta_ledger: Ledger;
  beta_asset: Asset;
  alpha_expiry: number;
  beta_expiry: number;
  alpha_ledger_refund_identity?: string;
  beta_ledger_redeem_identity?: string;
  peer: Peer;
}
```

Since comit-network/comit-rs#1602 was merged, `cnd`'s HTTP API allows for sending swap requests without specifying expiries (it defaults to conservative values: 24 hours for alpha and 12 hours for beta).

## Goal

Make expiries optional at the SDK level, so that developers can delegate the work of figuring out expiries to `cnd`.

## Recommendation

Make `alpha_expiry` and `beta_expiry` optional:

```ts
export interface SwapRequest {
  alpha_ledger: Ledger;
  alpha_asset: Asset;
  beta_ledger: Ledger;
  beta_asset: Asset;
  alpha_expiry: number?;
  beta_expiry: number?;
  alpha_ledger_refund_identity?: string;
  beta_ledger_redeem_identity?: string;
  peer: Peer;
}
```

luckysori:
`cnd` has now been able to derive expiries for you [for a while](https://github.com/comit-network/comit-rs/pull/1602). Mind you this is still not released, but it will work with `comit-rs@master`.

Fixes #52.

---

As a side note, it is extremely painful to test this using `create-comit-app`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 node_modules/
 
+dist/
+
 .idea/
 
 # This is a library
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,5 @@
 node_modules/
 
-dist/
-
 .idea/
 
 # This is a library
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Make alpha and beta expiries optional in `SwapRequest`.
+
 ## [0.6.0] - 2019-11-25
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "@types/express": "^4.17.2",
     "@types/jest": "^24.0.18",
     "@types/moment": "^2.13.0",
+    "@types/node": "^12.12.12",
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -1,5 +1,7 @@
 node_modules/
 
+dist/
```
D4nte: 📣 `npm publish` does not include files which are git ignored, hence you should not do that.
Feel free to double check by running `npm publish --dry` and check the list of files that it would publish.

luckysori: I believe you. It's annoying that the `dist` directory is always generated and untracked after building though :(

luckysori:
> If the cnd needed is not released then this cannot be merged and should be a draft PR... no?

Is the SDK tied to a specific release of 'cnd' or just supposed to be up-to-date with 'cnd@master'?

thomaseizinger:
> If the cnd needed is not released then this cannot be merged and should be a draft PR... no?

As long as we release it as a breaking change, that should not matter :)

PR #65, opened 2019-11-26
Negotiation follow up

D4nte:
Follow-up of comments on #60 and comit-network/create-comit-app#208

### DIFF
--- a/src/negotiation/execution_params.ts
+++ b/src/negotiation/execution_params.ts
@@ -40,7 +40,9 @@ export function validateExecutionParams(
       // For mainnet, we expect 24 hours and 48 hours expiry.
       // Checking 23 hours/47 hours cause we're lazy and it's still good enough
       return (
-        relativeAlphaExpiry > 47 * 60 * 60 && relativeBetaExpiry > 23 * 60 * 60
+        relativeAlphaExpiry > 47 * 60 * 60 &&
+        relativeBetaExpiry > 23 * 60 * 60 &&
+        relativeAlphaExpiry > relativeBetaExpiry
       );
     }
     case NetworkType.AllTest: {
--- a/src/negotiation/execution_params.ts
+++ b/src/negotiation/execution_params.ts
@@ -37,11 +37,11 @@ export function validateExecutionParams(
 
   switch (getNetworkType(executionParams.ledgers)) {
     case NetworkType.AllMainnet: {
-      // For mainnet, we expect 24 hours and 48 hours expiry.
-      // Checking 23 hours/47 hours cause we're lazy and it's still good enough
+      // For mainnet, we expect 12 hours and 24 hours expiry.
+      // Checking 11 hours/23 hours cause we're lazy and it's still good enough
       return (
-        relativeAlphaExpiry > 47 * 60 * 60 &&
-        relativeBetaExpiry > 23 * 60 * 60 &&
+        relativeAlphaExpiry > 23 * 60 * 60 &&
+        relativeBetaExpiry > 11 * 60 * 60 &&
         relativeAlphaExpiry > relativeBetaExpiry
       );
     }
--- a/src/negotiation/execution_params.ts
+++ b/src/negotiation/execution_params.ts
@@ -10,7 +10,11 @@ export interface ExecutionParams {
   };
   alpha_expiry: number;
   beta_expiry: number;
-  ledgers: LedgerParams;
+  ledgers?: LedgerParams;
+}
+
+export function defaultLedgerParams(): LedgerParams {
+  return { bitcoin: { network: "mainnet" }, ethereum: { chain_id: 1 } };
 }
 
 interface LedgerParams {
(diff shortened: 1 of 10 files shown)


src/negotiation/order.ts
```diff
@@ -73,18 +82,46 @@ export function assetOrderToSwap(orderAsset: OrderAsset): Asset | undefined {
...
       name: token.type.toLowerCase(),
-      quantity: orderAsset.amount,
+      quantity: orderAsset.nominalAmount,
       token_contract: token.address
     };
   }
 
   return undefined;
 }
+function areAmountEqual(
```
da-kami:
```suggestion
function areAmountsEqual(
```

src/negotiation/order.ts
```diff
@@ -50,18 +51,26 @@ export function orderSwapAssetMatchesForMaker(
   if (isNative(orderAsset)) {
     return (
       swapAsset.name === orderAsset.asset &&
-      swapAsset.quantity === orderAsset.amount &&
+      areAmountEqual(
```
da-kami:
```suggestion
      areAmountsEqual(
```

src/negotiation/order.ts
```diff
@@ -50,18 +51,26 @@ export function orderSwapAssetMatchesForMaker(
...
   }
 
   if (swapLedger.name === "ethereum") {
     const token = getToken(orderAsset.asset);
     if (token) {
-      // TODO: It may make more sense to use eponymous unit in order instead of smallest unit (e.g. Use Ether instead of Wei
       return (
         swapAsset.name.toLowerCase() === token.type.toLowerCase() &&
-        swapAsset.quantity === orderAsset.amount
+        areAmountEqual(
```
da-kami:
```suggestion
        areAmountsEqual(
```

src/negotiation/execution_params.ts
```diff
@@ -26,6 +30,11 @@ export function validateExecutionParams(
   const now = moment().unix();
   const relativeAlphaExpiry = executionParams.alpha_expiry - now;
   const relativeBetaExpiry = executionParams.beta_expiry - now;
+
```
bonomat: Sorry that this feedback comes a bit late as it is not related to your PR but I have 3 points for the overall function `validateExecutionParams`:

1) the function validates if the two defined networks are of the same category and returns false otherwise: 
--> I personally would not enforce this without giving the user a possibility to disable this explicitly or at least stating so. 

2) imho the function is more about validating the expiries time and I would call it:
--> `validateExpiries`

3) There is a _bug_ on line 43: `relativeAlphaExpiry > 47 * 60 * 60 && relativeBetaExpiry > 23 * 60 * 60` 
--> it does not enforce the `alphaExpiry` to be greater then `betaExpiry` and that there is enough time in between. Imho this test should fail: 
```javascript
it("Invalidates mainnet execution parameters due to expiry", () => {
    const executionParams: ExecutionParams = {
      ...defaultExecutionParams(),
      alpha_expiry: moment().unix() + 24 * 60 * 60,
      beta_expiry: moment().unix() + 48 * 60 * 60,
    };

    expect(validateExecutionParams(executionParams)).toBeFalsy();
  });
```


--- 

Point 1 & 2 are not that important, point 3 should be fixed though. 
Maybe for now this would be good enough: 
```
relativeBetaExpiry > 23 * 60 * 60 && relativeAlphaExpiry > 2 * relativeBetaExpiry
```

D4nte: 👍
I'd either invite @da-kami to add that to the feedback we want. What do dev think about expiries?
Yes ultimately we want this to be flexible but my understanding is that for now, we set it.

src/negotiation/execution_params.ts
```diff
@@ -26,6 +30,11 @@ export function validateExecutionParams(
...
+  if (!executionParams.ledgers) {
```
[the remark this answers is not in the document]
D4nte: That is some advanced TypeScripting!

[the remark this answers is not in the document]
D4nte: Yeah, this whole executionParams thing can be better so I am going to leave it like that for now.

src/negotiation/order.ts
```diff
@@ -71,20 +80,74 @@ export function orderSwapAssetMatchesForMaker(
...
+      name,
+      quantity: quantity.toString(),
       token_contract: token.address
     };
   }
 
   return undefined;
 }
+
+function areAmountsEqual(
```
luckysori: :thinking: As a future improvement it could be nice to create a generic interface `Equals<T>` to be implemented on types such as `BitcoinAmount` and `EtherQuantity`.

src/negotiation/maker_negotiator.ts
```diff
@@ -51,14 +51,17 @@ export class MakerNegotiator {
 
   private async tryAcceptSwap(
     order: Order,
-    { timeout, tryInterval }: TryParams
+    { maxTimeoutSecs, tryIntervalSecs }: TryParams
```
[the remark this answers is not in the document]
luckysori: Same for `swap.ts`.

luckysori:
Added `no-mergify` so that you can fix the expiry values if you want.

src/negotiation/order.ts
D4nte: We don't have such types but I see your point :)

PR #66, opened 2019-11-26
Improve negotiation protocol by allowing swap retrieval by id

D4nte:
Implement this protocol negotiation improvement:

```
// This (finding swap using order details)
// should be simplified by retrieving via order id
// Or by doing an extra round trip:
// Taker -->(ask exec params)--> Maker
// Taker <--(exec params)<-- Maker
// Taker -->(send swap req)--> Taker cnd
// Taker -->(accept order, inc. swap id)--> Maker
// Maker -->(retrieve. check and accept with swap id)--> Maker cnd
// In any case, a matching logic to verify that the taker sent
// the correct parameter needs to happen but this verification
// step may not be done in the ComitClient
```

Also, fix auto-accept.

### DIFF
--- a/src/comitClient.ts
+++ b/src/comitClient.ts
@@ -1,7 +1,6 @@
 import { BitcoinWallet } from "./bitcoinWallet";
 import { Cnd, SwapRequest, SwapSubEntity } from "./cnd";
 import { EthereumWallet } from "./ethereumWallet";
-import { Order, orderSwapMatchesForMaker } from "./negotiation/order";
 import { Action, EmbeddedRepresentationSubEntity, Entity } from "./siren";
 import { Swap } from "./swap";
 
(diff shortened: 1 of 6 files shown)


src/negotiation/maker_negotiator.ts
```diff
@@ -29,44 +35,58 @@ export class MakerNegotiator {
...
 
-  public acceptOrder(
-    // @ts-ignore
-    order: Order
-  ): ExecutionParams | undefined {
+  public getExecutionParams(): ExecutionParams {
+    return this.executionParams;
+  }
+
+  public acceptOrder(swapId: string, order: Order) {
```
da-kami: ❓ I don't understand why the swap ID is back. If I don't misunderstand something, we changed this to be by Order on purpose. The swap ID is an execution parameter, that the maker comes up with after the taker has taken the order. I am not we keep the boundaries correctly here. 
The `ComitClient` now exposes `retrieveSwapById(swapId: string)` again. If we call `acceptOrder(swapId: string, order: Order)` this implies we have to remember the `swapId` in the application, not just the Order. Is this what we want?

src/negotiation/maker_negotiator.ts
```diff
@@ -29,44 +35,58 @@ export class MakerNegotiator {
...
+    swapId: string,
+    order: Order,
+    tryIntervalSecs: number
+  ) {
     while (true) {
-      await sleep(tryInterval);
+      await sleep(tryIntervalSecs * 1000);
 
-      const swap = await this.comitClient.retrieveSwapByOrder(order);
+      const swap = await this.comitClient.retrieveSwapById(swapId);
```
da-kami: A more general statement, because I just stumbled about it: 
I feel the negotiation should be better encapsulated in the SDK. I feel the `MakerNegotiator` should not know about the `ComitClient`, but the swap should be fetched and then passed from the outside. If the `MakerNegotiator` knows about the `ComitClient` that couples the negotiation and execution tightly. If we later plan to take the execution into a separate module, would the common denominator be the `ComitClient`?

You don't have to give me a lengthy reply @D4nte - we can discuss this tomorrow, just wanted to write down my thoughts in the right place.

src/negotiation/order.ts
```diff
@@ -73,18 +82,46 @@ export function assetOrderToSwap(orderAsset: OrderAsset): Asset | undefined {
...
       name: token.type.toLowerCase(),
-      quantity: orderAsset.amount,
+      quantity: orderAsset.nominalAmount,
       token_contract: token.address
     };
   }
 
   return undefined;
 }
+function areAmountEqual(
```
da-kami: seems this was branched from `negotation-follow-up` but this PR goes against `master`...

src/negotiation/maker_negotiator.ts
D4nte: This as per the explanation I gave you and Philipp about a negotiation protocol improvement when using the white board.

src/negotiation/maker_negotiator.ts
D4nte: This is designed as per @bonomat's original request to automate the swap request and acceptance.

src/negotiation/order.ts
D4nte: Please see the PR's description.

src/negotiation/maker_negotiator.ts
da-kami: Sorry, guess I forgot about that discussion 😅 - Go ahead then 😃

src/negotiation/maker_negotiator.ts
da-kami: I will stop putting such comments in, guess we have to roll with something and think about the bigger picture after we have feedback :)

src/negotiation/order.ts
da-kami: Touché 😛

src/comitClient.ts
```diff
@@ -55,15 +54,20 @@ export class ComitClient {
...
-        return (
-          swap.actions &&
-          !!swap.actions.find((action: Action) => {
-            return action.name === "fund" || action.name === "redeem";
-          })
-        );
+      .filter((swap: SwapSubEntity) => {
+        if (swap.properties && swap.properties.status === "IN_PROGRESS") {
+          if (swap.actions && swap.actions.length !== 0) {
+            return !!swap.actions.find((action: Action) => {
```
bonomat: I think there is some complexity in this if statements :D 

Can you simplify this a bit please, e.g. according the return early approach: 

```
if (!swap.properties) {
  return false;
}
if (swap.properties.status !== "IN_PROGRESS") {
  return false
}
...
```

This would improve the readability of this function.

D4nte: Resolving in #76

PR #70, opened 2019-12-04
SDK fixes

luckysori:
Preparing to release 0.7.1.


src/index.ts
```diff
@@ -23,3 +23,22 @@ export { BigNumber } from "ethers/utils";
 export { ComitClient } from "./comitClient";
 
 export { Swap } from "./swap";
+
+export {
```
da-kami: as discussed: not best practice to do this, but given that we don't know what parts the users want to control yet I am happy with exposing everything.

PR #72, opened 2019-12-06
treat conversion of Bitcoin different than Ethereum in from Nominal

da-kami:
Bitcoin has to allow a fraction.
It should not use Ethers BigNumber for the conversion.

### DIFF
--- a/src/negotiation/order.ts
+++ b/src/negotiation/order.ts
@@ -124,30 +124,31 @@ function areAmountsEqual(
   return amount.eq(new BigNumber(unitAmount));
 }
 
-function fromNominal(
+export function fromNominal(
   asset: string,
   nominalAmount: string,
   token?: Token
 ): BigNumber | undefined {
-  const nominal = new BigNumber(nominalAmount);
-
-  let decimals = 0;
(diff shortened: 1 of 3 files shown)


src/negotiation/order.ts
```diff
@@ -124,30 +124,31 @@ function areAmountsEqual(
...
-    case "ether": {
-      decimals = 18;
-      break;
-    }
-    default: {
-      if (token) {
-        decimals = token.decimals;
-      } else {
-        return undefined;
+  if (asset === "bitcoin") {
```
D4nte: Fractional amount should also be supported for Ether and ERC20 tokens.

src/negotiation/order.ts
```diff
@@ -124,30 +124,31 @@ function areAmountsEqual(
...
+    const nominal = parseFloat(nominalAmount);
+    const actual = nominal * 100000000;
+    return new BigNumber(actual);
+  } else {
+    const nominal = new BigNumber(nominalAmount);
```
D4nte: Should alwyas do `parseFloat`

PR #79, opened 2019-12-11
InMemoryBitcoinWallet

bonomat:
With this PR I extracted an interface for the `BitcoinWallet` so that a developer can provide her own bitcoin wallet to the actor. 
This is also needed if you want to provide a specifically configured wallet for Testnet.
I also renamed the `BitcoinWallet` to `InMemoryBitcoinWallet` to indicate that this is for in-memory use only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **Breaking change** of the Bitcoin wallet. Made the provided wallet to an in-memory wallet only and extracted an interface so that developers can implement their own wallet.
+
 ## [0.7.2] - 2019-12-05
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- `Breaking API` of the Bitcoin wallet. Made the provided wallet to an in-memory wallet only and extracted an interface so that developers can implement their own wallet.
```
D4nte:
```suggestion
- **Breaking API** of the Bitcoin wallet. Made the provided wallet to an in-memory wallet only and extracted an interface so that developers can implement their own wallet.
```

src/bitcoinWallet.ts
```diff
@@ -1,7 +1,26 @@
 import { Amount, Chain, Network, Pool, TX, WalletDB } from "bcoin";
 import Logger from "blgr";
 
-export class BitcoinWallet {
+export interface BitcoinWallet {
+  getAddress(): Promise<string>;
+
+  getBalance(): Promise<any>;
```
D4nte:
```suggestion
  getBalance(): Promise<number>;
```

thomaseizinger:
> I'll approve but I am sure @thomaseizinger will say we can somehow do better than using abstractions :)

I don't mind the abstractions :)
This is exactly why I was alwaya saying, if we _need_ an interface, it is easily extracted! I am not sure what the 11 commits are doing here but I am totally fine with the resulting code change :)

PR #83, opened 2020-01-11
GetAmount in BitcoinWallet returns a number

bonomat:


bonomat:
> Looks good but it seems there is an error.

@D4nte : I'd appreciate if you have a spare minute to look into this. Imho this error is not related to my single line code change (and it works on my machine) :S

D4nte:
> > Looks good but it seems there is an error.
> 
> @D4nte : I'd appreciate if you have a spare minute to look into this. Imho this error is not related to my single line code change (and it works on my machine) :S

Please always delete your `yarn.lock` file locally before investigating CI issues :)

PR #84, opened 2020-01-12
Fix abi import

D4nte:
Looks like ethers interface changed but it went
through the net.

### DIFF
--- a/src/ethereumWallet.ts
+++ b/src/ethereumWallet.ts
@@ -4,9 +4,9 @@ import { TransactionRequest } from "ethers/providers";
 import {
   Arrayish,
   BigNumber as BigNumberEthers,
-  ParamType,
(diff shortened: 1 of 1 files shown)


D4nte:
Ok so first Master build is not triggered regularly. I wonder if we can do that with GitHub: https://github.com/comit-network/comit-js-sdk/runs/351614627

D4nte:
> Shall we investigate why this happened?

ethers changed their API from 4.0.39 to 4.0.41.... 
https://github.com/ethers-io/ethers.js/compare/v4.0.41...master

I don't know how can this be helped as we do not check in the `yarn.lock` file.
I think in this case the best we can do is regular build on master and investigate if it breaks.

@thomaseizinger any other idea?

PR #100, opened 2020-01-30
Hide MakerHttpApi, add getUrl method

D4nte:
See comit-network/create-comit-app#371

### DIFF
--- a/src/negotiation/maker_negotiator.ts
+++ b/src/negotiation/maker_negotiator.ts
@@ -63,8 +63,8 @@ export class MakerNegotiator {
     return this.makerhttpApi.getUrl();
   }
 
-  public listen(port: number) {
-    return this.makerhttpApi.listen(port);
+  public listen(port: number, hostname?: string) {
+    return this.makerhttpApi.listen(port, hostname);
   }
 
   private tryAcceptSwap(
@@ -132,7 +132,7 @@ class MakerHttpApi {
     this.server = undefined;
   }
 
-  public listen(port: number) {
+  public async listen(port: number, hostname?: string) {
     const app = express();
 
     app.use(express.json());
@@ -167,9 +167,29 @@ class MakerHttpApi {
       }
     });
 
-    this.server = app.listen(port, () =>
-      console.log(`Maker's Negotiation Service is listening on port ${port}.`)
-    );
+    if (hostname) {
+      this.server = app.listen(port, hostname, () =>
+        console.log(
+          `Maker's Negotiation Service is listening on ${hostname}:${port}.`
+        )
+      );
+    } else {
+      this.server = app.listen(port, () =>
+        console.log(`Maker's Negotiation Service is listening on port ${port}.`)
+      );
+    }
+
+    // Waiting for the server to start
+    let i = 0;
+    while (!this.server.listening && i < 10) {
+      await sleep(100);
+      i++;
+    }
+    if (!this.server.listening) {
+      throw new Error(
+        `There was an error starting the http server, is ${port} already in used?`
+      );
+    }
   }
 
   public getUrl(): undefined | string {
--- a/src/negotiation/maker_negotiator.ts
+++ b/src/negotiation/maker_negotiator.ts
@@ -1,4 +1,5 @@
 import express from "express";
+import * as http from "http";
 import { ComitClient } from "../comitClient";
 import { sleep, timeoutPromise, TryParams } from "../timeout_promise";
 import { ExecutionParams } from "./execution_params";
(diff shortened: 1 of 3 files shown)


src/negotiation/maker_negotiator.ts
```diff
@@ -120,28 +151,40 @@ export class MakerHttpApi {
...
+    this.server = app.listen(port, () =>
       console.log(`Maker's Negotiation Service is listening on port ${port}.`)
     );
   }
 
-  public addOrder(order: Order) {
-    this.maker.addOrder(order);
+  public getUrl(): undefined | string {
+    if (this.server) {
+      const addr = this.server.address();
```
bonomat: Do you know what address this is returning? 
I could imagine that this would return the address it is listening on, e.g .`0.0.0.0` on default. 

An alternative approach would be to explicitly define the listening hostname as described here: 
https://nodejs.org/api/synopsis.html

e.g. 
```
const http = require('http');

const hostname = '127.0.0.1';
const port = 3000;

const server = http.createServer((req, res) => {
  res.statusCode = 200;
  res.setHeader('Content-Type', 'text/plain');
  res.end('Hello, World!\n');
});

server.listen(port, hostname, () => {
  console.log(`Server running at http://${hostname}:${port}/`);
});
```

D4nte: It returns IPV6 localhost at the moment.
I think the ultimate approach would be to either:
- Use libp2p
- Allow user to configure what is the external ip of the maker

D4nte: Interestingly enought, `app.listen(3218, "127.0.0.1)` does not work, making my life more complicated.

D4nte: I made something, have a look please.

bonomat: > It returns IPV6 localhost at the moment.
> I think the ultimate approach would be to either:
> 
> * Use libp2p
> * Allow user to configure what is the external ip of the maker
> 
> Even if we allow the user to pass an ip to the listen function (and I can do that easily) this ultimately may not help behind a nat etc.

libp2p probably will make it worse first as people are not used to the multi-address format. 

Feel free to keep it as it is if it works in a useful way. 
If we only optimize for these two lines of code then we probably don't need this function at all:

https://github.com/comit-network/create-comit-app/blob/35f956e0ffd0e47be13e58cf85e924988ec6175c/create/new_project/examples/btc_eth/src/maker.ts#L85-L88

We should keep in mind that the sdk should be useful outside of the examples. 
I have the feelings we trim the sdk to make life easy for the examples but forget about real world usability. 

What do you think?

p.s. sorry, I think we should have discussed this earlier.

D4nte: > > It returns IPV6 localhost at the moment.
> > I think the ultimate approach would be to either:
> > 
> > * Use libp2p
> > * Allow user to configure what is the external ip of the maker
> > 
> > Even if we allow the user to pass an ip to the listen function (and I can do that easily) this ultimately may not help behind a nat etc.
> 
> libp2p probably will make it worse first as people are not used to the multi-address format.
> 
> Feel free to keep it as it is if it works in a useful way.
> If we only optimize for these two lines of code then we probably don't need this function at all:
> 
> [comit-network/create-comit-app:create/new_project/examples/btc_eth/src/maker.ts@`35f956e`#L85-L88](https://github.com/comit-network/create-comit-app/blob/35f956e0ffd0e47be13e58cf85e924988ec6175c/create/new_project/examples/btc_eth/src/maker.ts#L85-L88)
> 
> We should keep in mind that the sdk should be useful outside of the examples.
> I have the feelings we trim the sdk to make life easy for the examples but forget about real world usability.
> 
> What do you think?
> 
> p.s. sorry, I think we should have discussed this earlier.

Did you look at my last commit? https://github.com/comit-network/comit-js-sdk/pull/100/commits/d0d12011831d7d27ce3396a1c73a66c82e0d74cb

bonomat: Yes, saw it. Looks good.

What do you think about my point?

D4nte: I am not sure I understand your point.
I feel like `getUrl()` has some usefulness outside the sdk.

bonomat: Let's not over discuss this topic, it is such a small change.

It's good if you see the usefulness outside of the examples. That's what I wanted to point out. 

So let's keep it as you propose.

src/negotiation/maker_negotiator.ts
```diff
@@ -120,28 +151,60 @@ export class MakerHttpApi {
...
+
+    // Waiting for the server to start
+    let i = 0;
+    while (!this.server.listening && i < 10) {
+      await sleep(100);
+      i++;
+    }
+    if (!this.server.listening) {
+      throw new Error(
+        `There was an error starting the http server, is ${port} already in used?`
```
da-kami:
```suggestion
        `There was an error starting the http server, is ${port} already in use?`
```

Issue #97, opened 2020-01-29
Test external calls (cnd.ts)


da-kami:
## Problem

If cnd goes down or a library throw an exception, it's finished.

## Goal

Be robust and maintain it by reviewing how to make the sdk robust to runtime failures (cnd call times out, functions throwing errors).

## Recommendation

1. Identify _external_ calls especially in cnd.ts (e.g, calls to external libraries, network, io, etc).
1. Protect those calls (`try` ?)
1. Test these protections: Jest has good way to do some mockings.

D4nte:
Have a look at #111 as it can be one way to do this issue.

PR #111, opened 2020-02-10
Improve try* error handling by using results

D4nte:
Got a bit frustrated when creating e2e tests due to the poor error output.

Might help with #97.


D4nte:
Hi @yosriady what do you think of this proposed evolution?

I am trying to ensure the lib consumer knows why something might have fail on the SDK side, which is not always clear.

yosriady:
Do you have a unit test that shows the new error handling flow in action?

D4nte:
To be done properly with #97

PR #114, opened 2020-02-13
Update Typedoc markdown generation command

yosriady:
Minor changes around API doc generation. Specifically:

- Specify additional flags for Typedoc markdown generation
- Fix broken img tag (Please remember to close `` tags, orphaned HTML tags will break some Markdown compilers.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <a href="https://comit.network">
-<img src="logo.svg" height="120px">
+  <img src="logo.svg" height="120px" />
 </a>
 
 ---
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,5 +1,5 @@
 <a href="https://comit.network">
-<img src="logo.svg" height="120px">
+  <img src="logo.svg" height="120px" />
```
da-kami: Oh, my bad 😅

da-kami: added `alt` as well, will update the other READMEs accordingly, so we don't run into this problem at a different point again.

PR #116, opened 2020-02-14
Improve `TakerNegotiator` interface

### LINKED ISSUE #89: Taker's order criteria
## Problem

To get an order, the taker needs to go through several hoops:
1. Get orders from maker by doing `TakerNegotiator.getOrderByTradingPair("ETH-BTC")`
1. Do manual sanity checks on the order: https://github.com/comit-network/create-comit-app/blob/e2718a60bf7319e2504bb355c0406b00da0ae055/create/new_project/examples/separate_apps/src/taker.ts#L49
1. Manually check that the rate is acceptable: https://github.com/comit-network/create-comit-app/blob/e2718a60bf7319e2504bb355c0406b00da0ae055/create/new_project/examples/separate_apps/src/taker.ts#L67
1. Take the order using `TakerNegotiator.takeOrder(order)`: https://github.com/comit-network/create-comit-app/blob/e2718a60bf7319e2504bb355c0406b00da0ae055/create/new_project/examples/separate_apps/src/taker.ts#L73

The downsides are:
- The taker needs to specify the criteria in several steps (first pair, then check itself)
- Validation of the order is manual, a dev would expect to be able to do `Order.validate()` or even only return valid orders from the `TakerNegotiator.getOrderByTradingPair` call by filtering
- The filtering is quite manual and forces the dev to work through the properties of the order

## Goal

Provide a straightforward, function based, API to the developer. Alleviate the work.

## Recommendation

Re-think the API describe above to simplify it.
A better flow would be:

```
const criteria = {
	buy: "BTC",
	sell: "ETH",
	btcAmount: 0.5,
	allowOffersWithLesserAmount: false,
	maximumRate: 10
}

// Probably revise the `TakerNegotiator` terminology
const order = await takerNegotiator.getOffer(criteria);

if (!order.isValid()) { // This is an optional call and could/should be done already by `getOffer`
	// Ban this maker
}

if (!order.matches()) {  // This is an optional call and could/should be done already by `getOffer`
	// Tell the user why we cannot find an appropriate order
	console.log("Order found but does not match criteria", order.matchFailReason());
	// -> "Order found but does not match criteria { reason: 'maximumRate exceeded', expected: 10, actual: 15 }"
	// -> "Order found but does not match criteria { reason: 'lesser amount provided and not allowed', expected: 0.5, actual: 0.1 }"
}

const swap = await order.take();
```

D4nte:
This test demonstrates how to use the `TakerNegotiator`.
Note that the last part `await order.take()` is not tested as it involves creating a few mock. Not yet sure whether I should do it.

https://github.com/comit-network/comit-js-sdk/blob/f6d5cf070ac1d0c063d6afa4aca112ef16437ecc/src/negotiation/taker_negotiator.spec.ts#L25

Usage in the example drafted here: https://github.com/comit-network/create-comit-app/pull/440

Resolves #89


src/negotiation/order.ts
```diff
@@ -16,8 +16,93 @@ export interface OrderAsset {
...
+    }
+
+    const askAmount = new BigNumber(this.orderParams.ask.nominalAmount, 10);
+    const bidAmount = new BigNumber(this.orderParams.bid.nominalAmount, 10);
+
+    return !askAmount.isNaN() && !bidAmount.isNaN();
+  }
+
+  public async take(): Promise<Swap | undefined> {
+    if (this.matches() && this.isValid()) {
```
bonomat: I would turn the match around and check for validity first. Also, you can call `assetsMatch` directly.

```suggestion
    if (this.isValid() && this.assetsMatch()) {
```

src/negotiation/order.ts
```diff
@@ -152,3 +237,29 @@ export function fromNominal(
   }
   return new BigNumber(10).pow(decimals).times(nominalAmount);
 }
+
+export function orderParamsToTradingPair(orderParams: OrderParams): string {
```
bonomat: Why don’t you return an array or tuples or an object?

src/negotiation/taker_negotiator.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+  mockGetOrderByTradingPair,
+  // @ts-ignore
+  mockGetExecutionParams
+} from "./maker_client";
+
+jest.mock("../comit_client");
+jest.mock("./maker_client");
+
+const defaultCnd = new Cnd("");
+const defaultComitClient = new ComitClient(defaultCnd);
```
bonomat: If this is the reason for making `bitcoinWallet` and `ethereumWallet` optional, then just create two empty mocks here for those two.

src/actor.ts
```diff
@@ -24,7 +24,9 @@ export async function createActor(
     .getPeerListenAddresses()
     .then(addresses => addresses[0]);
 
-  const comitClient = new ComitClient(bitcoinWallet, ethereumWallet, cnd);
+  const comitClient = new ComitClient(cnd);
+  comitClient.ethereumWallet = ethereumWallet;
```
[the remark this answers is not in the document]
D4nte: Please check commit message.

D4nte:
> If I see it correctly then you had to make `bitcoinWallet` and `ethereumWallet` optional and expose these two fields publicly just to facilitate the tests.
> I think we should try to avoid this.

We won't need a Bitcoin Wallet if ones only wants to do LNBTC-ERC20 swaps. Hence I believe it is a move in the right direction.

src/negotiation/order.ts
D4nte: This needs to be a primitive type such as string or integer so it can be used as key for the hasmap and in the url of the REST API of makernegotiator

bonomat:
> We won't need a Bitcoin Wallet if ones only wants to do LNBTC-ERC20 swaps. Hence I believe it is a move in the right direction.

Ok, valid point.

src/actor.ts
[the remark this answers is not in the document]
bonomat: I think it is more efficient to answer instead of referring to a commit message. Saves time for the both of us.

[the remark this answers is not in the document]
D4nte: It is actually common practice in the team so I'd like to invite you to mention it at the next retro please.

src/negotiation/order.ts
```diff
@@ -16,8 +16,93 @@ export interface OrderAsset {
...
+  ) {}
+
+  public matches(): boolean {
+    return this.assetsMatch();
+  }
+
+  public isValid(): boolean {
+    if (
+      !(
+        this.orderParams.ask.ledger &&
```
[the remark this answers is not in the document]
D4nte: The `OrderParams` is an `Interface` so no, I can't implement `isValid()` on it (or at least I don't know how to do it in TS).

Also, I am keeping it an `Interface` and not a `Class` in purpose.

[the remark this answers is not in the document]
D4nte: Yet, I will make a separate function.

src/actor.ts
```diff
@@ -24,7 +24,9 @@ export async function createActor(
...
+  comitClient.bitcoinWallet = bitcoinWallet;
```
bonomat: Instead of exposing these fields publicly you could apply the _builder pattern_. This would allow for something like this: 

```
const comitClient = new ComitClient(cnd)
    .withEthereumWallet(ethereumWallet)
    .withBitcoinWallet(bitcoinWallet)
    ....;
```

Let me know what you think and I can provide a small PR once your changes are in.

src/negotiation/order.ts
```diff
@@ -1,9 +1,9 @@
 import { BigNumber } from "bignumber.js";
 import { Asset, Ledger, SwapProperties } from "../cnd";
+import { Swap } from "../swap";
 import { getToken, Token } from "../tokens/tokens";
 
-export interface Order {
-  tradingPair: string;
+export interface OrderParams {
```
da-kami: 🙃 I find the design of the `Order` a bit confusing. From code

```
export class Order {
  constructor(
    public readonly orderParams: OrderParams,
    public readonly criteria: TakerCriteria,
    public readonly takeOrder: (
      orderParams: OrderParams
    ) => Promise<Swap | undefined>
  ) {}
```

So the order consists of `OrderParams` and a `TakerCriteria`, why is the abstraction of `OrderParams` needed? What exactly is a `TakerCriteria`?

So it is cleaner to separate it from the `Order`? I feel there is quite a lot of classes now and it's not immediately understandable why and what they are for. 
Motivation for writing this: Started to write comments for code documentation and it's not that easy to explain what is going on. What is currently there works, but I wonder if we could design it so it is easier to understand. 

Can we simplify this by integrating the `OrderParams` into the `Order`? Is this abstraction really needed?

src/negotiation/maker_client.ts
```diff
@@ -0,0 +1,31 @@
...
+  public async getExecutionParams(
+    order: OrderParams
+  ): Promise<ExecutionParams> {
+    const response = await axios.get(
+      `${this.makerUrl}orders/${order.id}/executionParams`
+    );
+    return response.data;
+  }
+
+  public async takeOrder(order: OrderParams, swapId: string) {
```
da-kami: ❓ How come we already have a `swapId` at this point? Aren't we supposed to get the `swapId` from the maker **after** taking the order?

da-kami:
Overall: Like the approach towards tests, but don't understand why certain design decisions were made. Would be cool if could go over this with me Franck and explain it (Would help for creating good documentation).

src/actor.ts
D4nte: Please hold, I am finishing off few things. I'll ping you.

src/negotiation/order.ts
D4nte: Please note that the lib consumer is not supposed to constructor `Order` by themselves.
They would only use `OrderParams` (a simple object) to add orders to the order book.

Interesting point about documentation because this means that the constructor of `Order` should not be documented for example.

src/negotiation/maker_client.ts
D4nte: This is the swap Id indeed. I have documented in f622aea0

D4nte:
> Overall: Like the approach towards tests, but don't understand why certain design decisions were made. Would be cool if could go over this with me Franck and explain it (Would help for creating good documentation).

Please review @da-kami I added documentation.

src/negotiation/order.ts
da-kami: I think I understand what you are doing, but I still find the naming a bit confusing. I don't have a perfect solution in mind yet, but maybe we should step back, take a whiteboard and design this outside the code. The way `Order` and `OrderParams` is introduced in this PR is a bit confusing for me. we have a lot of variables named `order` now, that are actually `OrderParams`. Maybe a simple rename can fix this, not sure.

src/negotiation/order.ts
```diff
@@ -16,8 +16,125 @@ export interface OrderAsset {
...
+
+export interface TakerCriteriaAsset {
+  ledger: string;
+  asset: string;
+  minNominalAmount?: string;
+  maxNominalAmount?: string;
+}
+
+/**
+ * @description An order handler for the taker. It has helper functions to facilitate the handler of an
```
da-kami: 🔧 `handler` for me implies, that this class **deals with** orders rather than **being** an `Order`.

src/negotiation/order.ts
```diff
@@ -16,8 +16,125 @@ export interface OrderAsset {
...
+ * order by a taker. This should only be created via `TakerNegotiatior.getOrder()` and should not be constructed from
```
da-kami: 🙃 If `TakerNegotiatior.getOrder()` creates an `Order` it should probably be renamed to `TakerNegotiatior.getNewOrder()`

src/negotiation/order.ts
```diff
@@ -16,8 +16,125 @@ export interface OrderAsset {
...
+ * scratch.
+ * @param orderParams - The raw parameters of the order.
+ * @param criteria - The criteria used to get this order.
+ * @param takeOrder - Function passed from the `TakerNegotiator` to coordinate calls to `cnd` and the maker to effectively
+ * take the order and start the atomic swap execution.
+ */
+export class Order {
+  /**
+   * @description **Note: This should not be used, `Order` should be created by using `TakerNegotiatior.getOrder()`
```
da-kami: Can one make the constructor private then and use `TakerNegotiatior.getOrder()` like a factory?

src/negotiation/order.ts
```diff
@@ -16,8 +16,125 @@ export interface OrderAsset {
...
+    return (
+      assetMatches(this.criteria.buy, this.orderParams.bid) &&
+      assetMatches(this.criteria.sell, this.orderParams.ask) &&
+      rateMatches(this.criteria, this.orderParams)
+    );
+  }
+
+  /**
+   * @description Check that the order is valid and safe. Ensure that all properties are set and that the expiries
+   * are relatively safe. It does not check whether the ledgers/assets are correct, this is done with `Order.matches()`.
```
da-kami:
```suggestion
   * are safe. It does not check whether the ledgers/assets are correct, this is done with `Order.matches()`.
```

Feels wrong to call something `relatively safe` in the context of an atomic swap.

src/negotiation/order.ts
```diff
@@ -16,8 +16,125 @@ export interface OrderAsset {
...
+   */
+  public isValid(): boolean {
+    return areOrderParamsValid(this.orderParams);
+  }
+
+  /**
+   * @description Initiate the swap execution and tells the maker that we are taking this order.
```
da-kami:
```suggestion
   * @description Initiates the swap execution and tells the maker that we are taking this order.
```

or remove the `s` on tells

src/negotiation/taker_negotiator.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        ledger: "ethereum",
+        asset: "ether"
+      }
+    };
+
+    // Ask the maker if there are any orders that matches the criteria
+    const order = await takerNegotiator.getOrder(takerCriteria);
+
+    // We expect the returned order to be valid, if not
+    // This maker may be broken or malicious
```
da-kami:
```suggestion
    // this maker may be broken or malicious.
```

🔧 Any particular reason why you start each line caps? Reads difficult if it is a sentence. (Note: same applies for comments below)

src/negotiation/taker_negotiator.ts
```diff
@@ -47,25 +53,37 @@ export class TakerNegotiator {
...
 
-  public async takeOrder(order: Order): Promise<Swap | undefined> {
-    const executionParams = await this.makerNegotiator.getExecutionParams(
-      order
+  /**
+   * @description Get an order from the maker based on specified criteria. Whatever is returned from the maker is
+   * returned here, even if it does not match the criteria or is invalid. Not all criteria are passed to the maker.
+   * If it is indeed invalid or mismatching it will not be possible to execute the order, however it gives the
+   * opportunity to the lib consumer to know that this maker returns invalid orders and the details of such order.
+   * @param criteria - The criteria of the order that we are looking for.
```
da-kami:
```suggestion
   * @param criteria - The criteria of the order to be requested from the maker.
```

src/negotiation/taker_negotiator.ts
```diff
@@ -47,25 +53,37 @@ export class TakerNegotiator {
...
     );
-    if (!executionParams) {
-      return;
-    }
 
+    return new Order(orderParams, criteria, this.execAndTakeOrder.bind(this));
+  }
+
+  /**
+   * @description **Note: This should not be used, `Order.take()` should be preferred.**
```
da-kami: 🔧 if not preferred, can we hide this more properly?

src/negotiation/maker_client.ts
```diff
@@ -0,0 +1,41 @@
...
+   * @description **Note: `Order.take()` should be preferred.**
+   * Tells the maker that we are taking this order and what is the swap id of the swap request sent for
+   * this order from taker cnd to maker cnd.
+   * @param orderId - The id of the order that is taken.
+   * @param swapId - To facilitate matching between swap requests (cnd<>cnd) and orders, the taker first tells their
+   * cnd to send a swap request to the maker's cnd, then use the unique swap id, known by both cnds, to tell the maker
+   * that it is taking the order and what swap request is being use for this order. While this facilitates the
+   * order<>swap matching logic on the maker's side, the maker still needs to double check the parameters of the swap.
+   */
+  public async takeOrder(orderId: string, swapId: string): Promise<void> {
```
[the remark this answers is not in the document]
D4nte: The lib consumer is not supposed to use this function but de other in the SDK need to use it!

src/negotiation/taker_negotiator.spec.ts
D4nte: I have no idea.

src/negotiation/taker_negotiator.ts
D4nte: Can you please propose a way to "hide it more properly"?

src/negotiation/maker_client.ts
```diff
@@ -0,0 +1,41 @@
...
+    );
+    return response.data;
+  }
+
+  /**
+   * @description **Note: `Order.take()` should be preferred.**
+   * Tells the maker that we are taking this order and what is the swap id of the swap request sent for
+   * this order from taker cnd to maker cnd.
+   * @param orderId - The id of the order that is taken.
+   * @param swapId - To facilitate matching between swap requests (cnd<>cnd) and orders, the taker first tells their
```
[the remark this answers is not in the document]
D4nte: The new sentence is actually incorrect.

src/negotiation/maker_client.ts
```diff
@@ -0,0 +1,41 @@
...
+   * cnd to send a swap request to the maker's cnd, then use the unique swap id, known by both cnds, to tell the maker
+   * that it is taking the order and what swap request is being use for this order. While this facilitates the
+   * order<>swap matching logic on the maker's side, the maker still needs to double check the parameters of the swap.
```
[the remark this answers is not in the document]
D4nte: Yes but if you read this comment you are building the taker so I think it'd be confusing to point to code to use on the maker side. Instead, we just need to document the maker code.

src/negotiation/order.ts
D4nte: I rephrased.

src/negotiation/order.ts
D4nte: Not sure I agree here as the order comes from the order book and hence is not "new" per se.

src/negotiation/order.ts
D4nte: If the constructor is private then **I**, a developer working on comit-sdk, cannot create `Order`. What you want is the TS equivalent Rust's `pub(crate)` which does not exist in TS as far as I know.

src/negotiation/order.ts
D4nte: I guess this is a sign that we need to make the check safer are we currently arbitrary use 24/48 hours

src/negotiation/maker_negotiator.ts
```diff
@@ -69,7 +83,7 @@ export class MakerNegotiator {
 
   private tryAcceptSwap(
     swapId: string,
-    order: Order,
+    order: OrderParams,
```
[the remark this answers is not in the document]
D4nte: all fixed

src/negotiation/order.ts
```diff
@@ -16,20 +16,147 @@ export interface OrderAsset {
...
+ * order by a taker. This should only be instantiated via `TakerNegotiatior.getOrder()` and should not be constructed from
+ * scratch.
+ * @param orderParams - The raw parameters of the order.
+ * @param criteria - The criteria used to get this order.
+ * @param takeOrder - Function passed from the `TakerNegotiator` to coordinate calls to `cnd` and the maker to effectively
+ * take the order and start the atomic swap execution.
+ */
+export class Order {
+  /**
+   * **Note: This should not be used, `Order` should be created by using `TakerNegotiatior.getOrder()`
```
da-kami:
```suggestion
```

Not a blocker, feel free to ignore.

PR #119, opened 2020-02-18
Add jest code coverage

### LINKED ISSUE #98: Test coverage tool for SDK
## Problem

We do not know how much of the SDK is tested

## Goal

Introduce and **document** a test coverage tool to measure progress of code quality.
This is to be used locally by the developer to assist with new code.

Then update #97 with what should be tested using the coverage.

## Recommendation

`jest --coverage`

bonomat:
Resolves #98 

Adds a code coverage report when tests are run. 

The code coverage can be found in: `./coverage/lcov-report/index.html ` and looks like this: 

image

It means that `src/negotiation` and `src/tokens` are well tested but most of the files are not.

I decided to not let a bot commit on the PR with the report and would like to motivate every dev to learn how to use this tool in a smart way. 

I think #97 is already well described. 
I'll create some more tickets around what can be tested.


D4nte:
Please add a "cov" script to the package.json

bonomat:
> Please add a "cov" script to the package.json
@D4nte: can can, would do you prefer running 
a) `jest`
b) `yarn cov`
? 
`jest` will generate the report by default now.

D4nte:
> > Please add a "cov" script to the package.json
> > @D4nte: can can, would do you prefer running
> > a) `jest`
> > b) `yarn cov` / `yarn run cov`
> > ?
> > `jest` will generate the report by default now.

Ah thanks, I did not realize `jest` would output it. Thanks.

PR #125, opened 2020-02-19
Use builder pattern

D4nte:


D4nte:
> Can the ComitClient function without the wallets?
> 
> Yes => this works,
> No => they should be mandatory ctor arguments.

- `ComitClient` will work with a `BitcoinWallet` and `EthereumWallet` wallet, without a `LightningWallet` very soon.
- `ComitClient` will work with a `LightningWallet` and `EthereumWallet` wallet, without a `BitcoinWallet` very soon.
- `ComitClient` will work with a `BitcoinWallet` and `GrinWallet` wallet, without an `EthereumWallet` one day.

PR #127, opened 2020-02-19
Add a check to ensure the dist/ folder is not ignored

D4nte:
Because otherwise the `dist/` folder is not included in the package.

Adding everyone in review so that nobody get surprised by the CI check.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -11,6 +11,16 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 10.x
+      - name: Check that dist folder is NOT gitignored
+        run: |
+          if [[ $(grep -c dist .gitignore) -eq 0 ]]
+            then
+              echo "dist/ is not gitignored"
+              exit 0
+            else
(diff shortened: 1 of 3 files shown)


thomaseizinger:
cc @yosriady :)

D4nte:
> cc @yosriady :)

Everyone include Yos should have gotten a notif as I requested them as reviewers :)

yosriady:
@D4nte I'm not sure this is the correct approach:

- **We don’t want to track compiled JavaScript files in our git repository.** In general, `dist/` should be in `.gitignore.` 

- **We do, however, want them sent to NPM when we publish.** The reason why adding `dist/` in our `.gitignore` makes the `npm publish` command ignore the directory is because: this package is missing an `.npmignore` file. 

[See how `.npmignore` works here](https://docs.npmjs.com/using-npm/developers.html#keeping-files-out-of-your-package). Based on this, what we should do is:

- Add `dist/` to `.gitignore`
- Create an `.npmignore` file.
- We wouldn't need to implement the CI check.

PR #133, opened 2020-02-21
Update prettier version

da-kami:
I had problems with the prettier version `"^1.18.2"` on my Mac, it would always touch the `siren.ts` file event though it should not. 
Updating version fixed this problem.

---

For reference, it would move the bracket on line `85` to line `84`:

```
      | "working-copy-of"
    );
/**
```

would become:

```
      | "working-copy-of");
/**
```

CI does not like that...


thomaseizinger:
Why do we have mergify updating the branch in this repo?

image

PR #135, opened 2020-02-24
Fix comment errors due to pointy brackets

da-kami:
Missed these two fixes when writing the docs, see https://github.com/comit-network/comit.network/pull/22#issuecomment-588528096


D4nte:
FYI, I triggered a manual check of the CLA just now.

PR #137, opened 2020-02-24
Replace mergify merging with bors

thomaseizinger:


D4nte:
> mergify does not like the config. Note that there is a config simulator availabe on mergify.io.

I fixed it for you.

PR #141, opened 2020-02-26
Add Lightning support for actions

### LINKED ISSUE #118: Extend the SDK to allow executing the new three actions
## Problem

During the execution of LN swaps, the client can face [three additional actions](https://github.com/comit-network/spikes/blob/master/0022-lnd-plan-of-attack.adoc#226-what-are-the-actions-we-need-to-emit-on-cnds-rest-api):

1. `lnd-add-hold-invoice`
2. `lnd-send-payment`
3. `lnd-settle-invoice`

## Goal

Make the SDK support LN swaps.

## Recommendation

Extend the `Swap` class with the necessary actions in https://github.com/comit-network/comit-js-sdk/blob/7bb5c658c2a456f5e4000ce801a3b7b66bda8964/src/swap.ts#L59-L88.

Action (1) and (3) are only available via gRPC (until hold-invoices are stabilized in lnd).
Meaning, whatever class we delegate these actions to, it will have to use gRPC to talk to lnd.

This includes the interaction with lnd. LightningLabs' Alex provide `ln-service` npm package, use it! In the e2e tests, there is types already defined in `e2e/types` for `ln-service`, copy them over.

The SDK should accept an `AuthenticatedLndGrpcApi` to connect to lnd and trigger the actions.
  
Check the lnd code in the e2e test and see if any unopinionated code should me moved over to the sdk.

Note: Test it! Most likely need to do some mocking.

D4nte:
This allows the use of comit-sdk to make current comit-rs e2e lnd tests pass.

See https://github.com/comit-network/comit-rs/pull/2107 for the e2e side.

Resolves #118.

### DIFF
--- a/src/cnd/cnd.ts
+++ b/src/cnd/cnd.ts
@@ -69,6 +69,24 @@ export interface EthereumCallContractPayload {
   network: string;
 }
 
+export interface LndSendPaymentPayload {
+  public_key: string;
+  amount: string;
+  secret_hash: string;
+  final_cltv_delta: number;
+}
+
+export interface LndAddHoldInvoicePayload {
+  amount: string;
+  secret_hash: string;
+  expiry: number;
+  memo: string;
+}
+
+export interface LndSettleInvoicePayload {
+  secret: string;
+}
+
 export type LedgerAction =
   | {
       type: "bitcoin-send-amount-to-address";
@@ -85,6 +103,18 @@ export type LedgerAction =
   | {
       type: "ethereum-call-contract";
       payload: EthereumCallContractPayload;
+    }
+  | {
+      type: "lnd-send-payment";
+      payload: LndSendPaymentPayload;
+    }
+  | {
+      type: "lnd-add-hold-invoice";
+      payload: LndAddHoldInvoicePayload;
+    }
+  | {
+      type: "lnd-settle-invoice";
+      payload: LndSettleInvoicePayload;
     }
   | {
       type: string;
@@ -274,7 +304,7 @@ export class Cnd {
    * @param body The body to set in the request. The design being not yet finalised it is optional and of type `any`
    * @return The location of the swap (href) as returned by the REST API in the location header.
    */
-  public async postHanEthereumEtherHalightLightningBitcoin(
+  public async createHanEthereumEtherHalightLightningBitcoin(
     body?: any
   ): Promise<string> {
     return axios
@@ -294,7 +324,7 @@ export class Cnd {
    * @param body The body to set in the request. The design being not yet finalised it is optional and of type `any`
    * @return The location of the swap (href) as returned by the REST API in the location header.
    */
-  public async postHerc20EthereumErc20HalightLightningBitcoin(
+  public async createHerc20EthereumErc20HalightLightningBitcoin(
     body?: any
   ): Promise<string> {
     return axios
@@ -314,7 +344,7 @@ export class Cnd {
    * @param body The body to set in the request. The design being not yet finalised it is optional and of type `any`
    * @return The location of the swap (href) as returned by the REST API in the location header.
    */
-  public async postHalightLightningBitcoinHanEthereumEther(
+  public async createHalightLightningBitcoinHanEthereumEther(
     body?: any
   ): Promise<string> {
     return axios
@@ -334,7 +364,7 @@ export class Cnd {
    * @param body The body to set in the request. The design being not yet finalised it is optional and of type `any`
    * @return The location of the swap (href) as returned by the REST API in the location header.
    */
-  public async postHalightLightningBitcoinHerc20EthereumErc20(
+  public async createHalightLightningBitcoinHerc20EthereumErc20(
     body?: any
   ): Promise<string> {
     return axios
--- a/src/lightning_wallet.ts
+++ b/src/lightning_wallet.ts
@@ -42,7 +42,7 @@ export class LightningWallet {
       dest: publicKeyBuf,
       amt: satAmount,
       paymentHash,
-      cltvLimit: finalCltvDelta
+      finalCltvDelta
     });
   }
 
@@ -53,19 +53,20 @@ export class LightningWallet {
     memo: string
   ): Promise<string> {
     const satAmountNum = parseInt(satAmount, 10);
-
+    const hash = Buffer.from(secretHash, "hex");
     return (
       await this.lnd.invoicesrpc.addHoldInvoice({
         value: satAmountNum,
-        hash: secretHash,
-        expiry,
-        memo
+        hash,
+        memo,
+        expiry
       })
     ).paymentRequest;
   }
 
   public async settleInvoice(secret: string): Promise<void> {
-    await this.lnd.invoicesrpc.settleInvoice({ preimage: secret });
+    const preimage = Buffer.from(secret, "hex");
+    await this.lnd.invoicesrpc.settleInvoice({ preimage });
   }
 
   public async newAddress(type: AddressType): Promise<string> {
--- a/src/lightning_wallet.ts
+++ b/src/lightning_wallet.ts
@@ -104,7 +104,9 @@ export class LightningWallet {
     const openChannel = this.lnd.lnrpc.openChannel(request);
 
     openChannel.on("error", (err: any) => {
-      console.log("Error encountered", err);
+      throw new Error(
+        `Error encountered for Open Channel: ${JSON.stringify(err)}`
+      );
     });
 
     const status: OpenStatusUpdate = await pEvent(openChannel, "data");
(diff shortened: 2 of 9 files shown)


src/lnd.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  public async openChannel(
+    request: OpenChannelRequest
+  ): Promise<{
+    transactionId: string;
+    transactionVout: number;
+  }> {
+    const openChannel = this.client.openChannel(request);
+
+    return openChannel.on("data", (status: ChannelOpenStatusUpdate) => {
```
D4nte: I am trying to return `transactionId` and `transactionVout` out of the `Lnd.openChannel()` method but I am not sure how I do that as I get the info from withing the `.on()` block.

src/lightning_wallet.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  public async settleInvoice(secret: string): Promise<void> {
+    await this.lnd.invoicesrpc.settleInvoice({ preimage: secret });
+  }
+
+  public async newAddress(type: AddressType): Promise<string> {
+    return (await this.lnd.lnrpc.newAddress({ type })).address;
+  }
+
+  public async confirmedChannelBalance(): Promise<string> {
```
thomaseizinger: Which channel is this? On an API of this abstraction level, should we not pass a channel identifier?

src/lightning_wallet.ts
```diff
@@ -0,0 +1,180 @@
...
+  public getLndP2pSocket(): string {
+    return `${this.p2pHost}:${this.p2pPort}`;
+  }
+
+  public async getInfo(): Promise<GetInfoResponse> {
+    return this.lnd.lnrpc.getInfo();
+  }
+
+  public async openChannel(
+    pubkey: string,
```
thomaseizinger:
```suggestion
    to: string,
```

src/lightning_wallet.ts
```diff
@@ -0,0 +1,180 @@
...
+    satAmount: number
+  ): Promise<Outpoint> {
+    const request = {
+      nodePubkey: Buffer.from(pubkey, "hex"),
+      localFundingAmount: satAmount.toString()
+    };
+    const openChannel = this.lnd.lnrpc.openChannel(request);
+
+    openChannel.on("error", (err: any) => {
```
thomaseizinger: :mega: This should be a failure of the function.

src/swap.ts
```diff
@@ -12,12 +13,40 @@ import { sleep, timeoutPromise, TryParams } from "./util/timeout_promise";
...
+  private lightningWallet?: LightningWallet;
+
   constructor(
-    private readonly bitcoinWallet: BitcoinWallet,
-    private readonly ethereumWallet: EthereumWallet,
     private readonly cnd: Cnd,
-    readonly self: string
-  ) {}
+    readonly self: string,
+    wallets?: {
```
thomaseizinger: I would create an interface for this and store the `Wallets` object directly inside. Then you can avoid the double optional thing and just default to an empty object inside the swap.
You may also take some inspiration from the `Wallets` class I created for the e2e tests: https://github.com/comit-network/comit-rs/blob/de1f4cae4d57f68fb3e9c2e8986b2b4dabdc4ac4/api_tests/lib/wallets/index.ts

thomaseizinger: It does exactly what you want here:

All wallets are optional but you can access them through an interface that gives you a wallet instance or throws an exception if it doesn't exist.

src/cnd/cnd.ts
```diff
@@ -69,6 +69,24 @@ export interface EthereumCallContractPayload {
...
+  amount: string;
+  secret_hash: string;
+  final_cltv_delta: number;
+}
+
+export interface LndAddHoldInvoicePayload {
+  amount: string;
+  secret_hash: string;
+  expiry: number;
+  memo: string;
```
D4nte: After further consideration, I believe the user should be the one to decide how to name the COMIT swap invoices and memo should not be part of the action.

@thomaseizinger thoughts?

src/cnd/cnd.ts
```diff
@@ -69,6 +69,24 @@ export interface EthereumCallContractPayload {
   network: string;
 }
 
+export interface LndSendPaymentPayload {
```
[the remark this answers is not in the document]
D4nte: Hum, I would actually question whether `network` is indeed in the others? If you think we should add it here I think a new issue would be better to avoid scope creep.

src/lightning_wallet.ts
```diff
@@ -0,0 +1,180 @@
...
+      tls: certPath,
+      macaroonPath,
+      server: `${lndHost}:${lndPort}`
+    });
+    return new LightningWallet(lnd, lndHost, lndPort);
+  }
+
+  public constructor(
+    public readonly lnd: Lnd,
+    private readonly p2pHost: string,
```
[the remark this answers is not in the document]
D4nte: I agree but after playing around I cannot find a way to do it. Please feel free to open a follow-up PR to show me how you'd do it.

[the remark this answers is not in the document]
D4nte: But thanks cause I found a bug thanks to this comment :)

src/lightning_wallet.ts
D4nte: The use of the API is correct here, you cannot specify the channel:

```
    /**
     * channelBalance returns the total funds available across all open channels in satoshis.
     */
    channelBalance(args?: {}): Promise<ChannelBalanceResponse>;
```

src/lightning_wallet.ts
D4nte: I do not understand the action you want me to take from this comment.

src/swap.ts
D4nte: I will do that in a follow-up PR because my OCD would trigger few actions:
- Create a `wallet/` folder in the SDK to contain this new class and existing wallets
- Copy over the `AllWallets` class from e2e
- Use the `AllWallets` class from SDK in e2e, and only extend it for the `initializeForLedger` bit.

src/cnd/cnd.ts
thomaseizinger: Yeah I am fine with that :)

src/swap.ts
thomaseizinger: > Copy over the AllWallets class from e2e

We should carefully consider that but maybe it is the right decision.

thomaseizinger:
Approved if you add a test that verifies we can throw exceptions in event handlers.

src/lightning_wallet.ts
D4nte: I would like to merge without testing that because my mastery of the jest framework for this specific test is just not good enough.

I can make the error happen, I just cannot assert it.

`lightning_wallet.spec.ts` content:
```
import { EventEmitter } from "events";
import { LightningWallet } from "./lightning_wallet";

jest.mock("./lnd");

describe("Lightning Wallet", () => {
  it("Propagate exception from event handler", async () => {

    const emitter = new EventEmitter();

    const mockOpenChannel = jest.fn().mockImplementation(() => {
      return emitter;
    });

    const lightningWallet = await LightningWallet.newInstance("", "","","");

    // @ts-ignore
    lightningWallet.lnd = {
      lnrpc: {openChannel: mockOpenChannel}
    };

    const openChannelPromise = lightningWallet.openChannel("", 1000);

    const expectPromise = expect(openChannelPromise).rejects.toMatch("Error encountered for Open Channel");
    emitter.emit("error");
    await expectPromise;
  });
});
```

Running the test results to:
```
Error: Error encountered for Open Channel: undefined

    at EventEmitter.openChannel.on.err (/Users/froyer/src/comit-js-sdk/src/lightning_wallet.ts:2685:15)
    at EventEmitter.emit (events.js:203:15)
    at /Users/froyer/src/comit-js-sdk/src/lightning_wallet.spec.ts:31:13
    at Generator.next (<anonymous>)
    at fulfilled (/Users/froyer/src/comit-js-sdk/src/lightning_wallet.spec.ts:5:58)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

Which is as expected.

Moreover, this test is not really about testing our code but it is about confirming that errors thrown in the callback of event listener are propagated. My patch above does prove it so I suggest we just move forward :)

src/cnd/cnd.ts
D4nte: Follow-up PR.

src/lightning_wallet.ts
thomaseizinger: > Moreover, this test is not really about testing our code but it is about confirming that errors thrown in the callback of event listener are propagated. My patch above does prove it so I suggest we just move forward :)

Agreed, let's move forward!

PR #145, opened 2020-03-04
Follow-up on #141

D4nte:


src/wallet/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  get bitcoin(): BitcoinWallet {
+    return this.getWalletForLedger("bitcoin") as BitcoinWallet;
+  }
+
+  get ethereum(): EthereumWallet {
+    return this.getWalletForLedger("ethereum") as EthereumWallet;
+  }
+
+  get lightning(): LightningWallet {
+    return this.getWalletForLedger("lightning") as LightningWallet;
```
thomaseizinger: I think you should be able to remove the casts here because `AllWallets[K]` is actually the type of the wallet and I think the optionality shouldn't leak out but maybe I am wrong.

Would be nice to get rid of this cast though :)

src/wallet/lightning.ts
```diff
@@ -134,6 +135,38 @@ export class LightningWallet {
       return { rHash: rHash.toString("hex"), paymentRequest };
     }
   }
+
+  public async assertLndDetails(
+    selfPublicKey: string,
+    chain: string,
+    network: string
+  ): Promise<void> {
+    const getinfo = await this.lnd.lnrpc.getInfo();
```
thomaseizinger: We could store that once during the initialization and then just reuse it for every action. Otherwise you will make that call all the time again :man_shrugging:

src/wallet/index.ts
D4nte: Somehow TS was not happy so I had to add the cast. However, incomit-rs e2e there is no cast! ☹️

src/wallet/lightning.ts
D4nte: Yes I went through the same thought process but then decided to not make assumptions or early optimisation hence the lack of caching.

PR #146, opened 2020-03-04
Export wallets

D4nte:


bonomat:
Just curious: what do we need the `AllWallets` interface and `Wallets ` class for?

D4nte:
> Just curious: what do we need the `AllWallets` interface and `Wallets ` class for?

`Wallets`: It's how it's internally stored in the `Swap` class. One might want to access them.
`AllWallets`: It's needed by the `Swap` constructor so best to expose it.

PR #151, opened 2020-03-05
Extend Cnd class with middleware for parsing problem responses

### LINKED ISSUE #122: Configure `axios` to handle problem responses
# Problem

The Axios client in the SDK does not handle  the media-type `application/problem+json` as returned by `cnd` if there is a problem.
We have some some behaviour where we would like to assert on message and status of a problem response (e.g. in the `bitcoin_high_fee` e2e test).

# Goal

Properly handle the media-type `application/problem+json` in Axios so we can assert on the problem.

# Proposed Solution

Note: @thomaseizinger recommended to check for plugins that do this, but I could not find any thus the recommendation is to add this to the axios config ourselves:

Check `transformResponse` here: https://github.com/axios/axios#request-config

Transform the `problem+json` response returned by `cnd` to return a proper `Problem` class that we can assert on it in some way.

Pseudo-code:
```
await expect(responsePromise).to.be.rejected
    .and.be.instanceOf("Problem")
    .and.have.property("name", value);
```

thomaseizinger:
We use axios interceptor functionality to parse responses of type 'application/problem+json' into a Problem class. This allows us to know about the fields present in the response and return a dedicated error to the user with useful information of what went wrong.

Resolves #122.

### DIFF
--- a/src/cnd/cnd.spec.ts
+++ b/src/cnd/cnd.spec.ts
@@ -1,0 +1,102 @@
+import nock = require("nock");
+import { Problem } from "./axios_rfc7807_middleware";
+import { Cnd } from "./cnd";
+
+it.each([
+  ["500", "Internal Server Error", "about:blank", null],
+  ["405", "Invalid action invocation", null, null],
+  [
+    "400",
+    "Swap not supported",
+    "about:blank",
+    "The requested combination of ledgers and assets is not supported."
+  ]
+])(
+  "should return a Problem if the response is application/problem+json",
+  (status, title, type, detail) =>
+    setup(async (interceptor, postSwap) => {
+      const statusCode = parseInt(status!, 10);
+
+      const scope = interceptor.reply(
+        statusCode,
+        JSON.stringify({ status, title, type, detail }),
+        {
+          "content-type": "application/problem+json"
+        }
+      );
+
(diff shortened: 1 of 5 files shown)


src/cnd/cnd.spec.ts
```diff
@@ -0,0 +1,102 @@
+import nock = require("nock");
+import { Problem } from "./axios_rfc7807_middleware";
+import { Cnd } from "./cnd";
+
+it.each([
+  ["500", "Internal Server Error", "about:blank", null],
+  ["405", "Invalid action invocation", null, null],
+  [
+    "400",
+    "Swap not supported",
```
D4nte: @thomaseizinger did you consider supporting a list of problems that cnd could return to allow the SDK consumer's to have better expectation and individual problem management?

thomaseizinger: Considered it yes but the way I would go about it is to actually have those problems point to a documentation using the `type` URL. See the RFC for details: https://tools.ietf.org/html/rfc7807#section-3.1

Those URLs can then be used by clients to differentiate between different errors.

I would say the current state is fine. Creating like a hard-coded list of problems would be over-engineered IMO and just create a maintenance burden.

PR #152, opened 2020-03-06
Document cnd errors, wrap cnd & wallet errors in specialised Error classes

D4nte:
_99% review_

I have selected the following approach for this PR after considering the feedback provided. I decided not to go with tuples as per @yosriady's [suggestion](https://github.com/comit-network/comit-js-sdk/pull/152#issuecomment-595603819) simply because it is a lot of work as I'd need to catch all failures to transform them to Promise resolution. It is also easy to miss. If we miss it would mean that the user end-up with a thrown error when they do not expect one.

We propagate by throwing when any error is encountered.
We try to _intercept_ any error to add more context to them (see `Problem` and `ChainError` classes)

The idea is that the lib consumer can catch those errors and manipulate them easily, as follow:

```
try {
	swap.redeem();
} catch (error) {
	if (error instanceOf ChainError) {
		// do something about the wallet
    } else if (error instanceOf Problem) {
      	// do something about cnd
    } else {
      	// display
	}
}
```

Resolves https://github.com/comit-network/comit-js-sdk/issues/97

### DIFF
--- a/src/cnd/axios_rfc7807_middleware.ts
+++ b/src/cnd/axios_rfc7807_middleware.ts
@@ -7,6 +7,7 @@ export class Problem extends Error {
   public readonly status?: number;
   public readonly detail?: string;
   public readonly instance?: string;
+  public readonly isProblem: boolean;
 
   constructor({ title, type, status, detail, instance }: ProblemMembers) {
     super(title);
@@ -15,6 +16,7 @@ export class Problem extends Error {
     this.detail = detail;
     this.instance = instance;
     this.title = title;
+    this.isProblem = true;
   }
 }
 
--- a/src/cnd/axios_rfc7807_middleware.ts
+++ b/src/cnd/axios_rfc7807_middleware.ts
@@ -7,7 +7,6 @@ export class Problem extends Error {
   public readonly status?: number;
   public readonly detail?: string;
   public readonly instance?: string;
-  public readonly isProblem: boolean;
 
   constructor({ title, type, status, detail, instance }: ProblemMembers) {
     super(title);
@@ -16,7 +15,6 @@ export class Problem extends Error {
     this.detail = detail;
     this.instance = instance;
     this.title = title;
-    this.isProblem = true;
   }
 }
 
--- a/src/cnd/cnd.ts
+++ b/src/cnd/cnd.ts
@@ -259,15 +259,27 @@ export class Cnd {
     );
   }
 
+  /**
+   * Get the peer id of the cnd node
+   *
+   * @returns Promise that resolves with the peer id of the cnd node,
+   * @throws A {@link Problem} from the cnd REST API or an {@link Error}.
+   */
   public async getPeerId(): Promise<string> {
     const info = await this.getInfo();
     if (!info.id) {
-      throw new Error("id field not present");
+      return Promise.reject(new Error("id field not present"));
     }
 
     return info.id;
   }
 
+  /**
+   * Get the address on which cnd is listening for peer-to-peer/COMIT messages.
+   *
+   * @returns An array of multiaddresses
+   * @throws A {@link Problem} from the cnd REST API or an {@link Error}.
+   */
   public async getPeerListenAddresses(): Promise<string[]> {
     const info = await this.getInfo();
     if (!info.listen_addresses) {
@@ -277,12 +289,25 @@ export class Cnd {
     return info.listen_addresses;
   }
 
+  /**
+   * Sends a swap request to cnd.
+   *
+   * @param swap The details of the swap to initiate.
+   * @returns The URL of the swap request on the cnd REST API.
+   * @throws A {@link Problem} from the cnd REST API or an {@link Error}.
+   */
   public async postSwap(swap: SwapRequest): Promise<string> {
     const response = await this.client.post("swaps/rfc003", swap);
 
     return response.headers.location;
   }
 
+  /**
+   * List the swaps handled by this cnd instance.
+   *
+   * @returns An Array of {@link SwapSubEntity}, which contains details of the swaps.
+   * @throws A {@link Problem} from the cnd REST API or an {@link Error}.
+   */
   public async getSwaps(): Promise<SwapSubEntity[]> {
     const response = await this.fetch("swaps");
     const entity = response.data as Entity;
@@ -290,10 +315,27 @@ export class Cnd {
     return entity.entities as SwapSubEntity[];
   }
 
+  /**
+   * Fetch data from the REST API.
+   *
+   * @param path The URL to GET.
+   * @returns The data returned by cnd.
+   * @throws A {@link Problem} from the cnd REST API or an {@link Error}.
+   */
   public fetch<T>(path: string): AxiosPromise<T> {
     return this.client.get(path);
   }
 
+  /**
+   * Proceed with an action request on the cnd REST API.
+   * This request could be a GET HTTP request, which only returns data needed to further proceed with the action,
+   * or a POST request, which is the action itself, to instruct cnd about a swap.
+   *
+   * @param action The action to perform.
+   * @param resolver A function that returns data needed to perform the action, this data is likely to be provided by a
+   * blockchain wallet or interface (e.g. wallet address).
+   * @throws A {@link Problem} from the cnd REST API, or {@link ChainError} if the blockchain wallet throws, or an {@link Error}.
+   */
   public async executeSirenAction(
     action: Action,
     resolver?: FieldValueResolverFn
--- a/src/cnd/cnd.ts
+++ b/src/cnd/cnd.ts
@@ -334,7 +334,7 @@ export class Cnd {
    * @param action The action to perform.
    * @param resolver A function that returns data needed to perform the action, this data is likely to be provided by a
    * blockchain wallet or interface (e.g. wallet address).
-   * @throws A {@link Problem} from the cnd REST API, or {@link ChainError} if the blockchain wallet throws, or an {@link Error}.
+   * @throws A {@link Problem} from the cnd REST API, or {@link WalletError} if the blockchain wallet throws, or an {@link Error}.
    */
   public async executeSirenAction(
     action: Action,
--- a/src/cnd/cnd.ts
+++ b/src/cnd/cnd.ts
@@ -328,8 +328,6 @@ export class Cnd {
 
   /**
    * Proceed with an action request on the cnd REST API.
-   * This request could be a GET HTTP request, which only returns data needed to further proceed with the action,
-   * or a POST request, which is the action itself, to instruct cnd about a swap.
    *
    * @param action The action to perform.
    * @param resolver A function that returns data needed to perform the action, this data is likely to be provided by a
--- a/src/cnd/cnd.ts
+++ b/src/cnd/cnd.ts
@@ -268,7 +268,7 @@ export class Cnd {
   public async getPeerId(): Promise<string> {
     const info = await this.getInfo();
     if (!info.id) {
-      return Promise.reject(new Error("id field not present"));
+      throw new Error("id field not present");
     }
 
     return info.id;
--- a/src/swap.spec.ts
+++ b/src/swap.spec.ts
@@ -1,6 +1,6 @@
 import nock = require("nock");
 import { Problem } from "./cnd/axios_rfc7807_middleware";
-import { Cnd, SwapDetails } from "./cnd/cnd";
+import { Cnd, LedgerAction, SwapDetails } from "./cnd/cnd";
 import { ChainError, Swap } from "./swap";
 
 describe("Swap", () => {
@@ -38,7 +38,7 @@ describe("Swap", () => {
     return scope;
   });
 
-  it("Throws a ChainError if a wallet returns an error when executing an action", async () => {
+  it("Throws a ChainError if a wallet returns an error when resolving fields", async () => {
     const basePath = "http://example.com";
     const cnd = new Cnd(basePath);
 
@@ -69,5 +69,62 @@ describe("Swap", () => {
     await expect(promise).rejects.toBeInstanceOf(ChainError);
 
     return scope;
+  });
+
+  it("Throws a ChainError if a wallet returns an error when executing an action", async () => {
+    const basePath = "http://example.com";
+    const cnd = new Cnd(basePath);
+
+    const selfPath = "/swap-id";
+
+    const swap = new Swap(cnd, `${basePath}/swap-id`, {});
+
+    const swapDetails: SwapDetails = {
+      actions: [
+        {
+          href: "/swap-id/deploy",
+          name: "deploy",
+          method: "GET"
+        }
+      ]
+    };
+
+    const swapScope = nock(basePath)
+      .get(selfPath)
+      .times(2)
+      .reply(200, JSON.stringify(swapDetails), {});
+
+    const actionPayload = {
+      type: "ethereum-deploy-contract",
+      payload: {
+        data: "",
+        amount: "",
+        gas_limit: "",
+        network: ""
+      }
+    };
+
+    const actionScope = nock(basePath)
+      .get("/swap-id/deploy")
+      .times(2)
+      .reply(200, JSON.stringify(actionPayload), {});
+
+    const tryParams = {
+      maxTimeoutSecs: 1,
+      tryIntervalSecs: 0.01
+    };
+
+    // To be sure we are testing the right thing, we verify this does not throws
+    const executeActionPromise = swap.tryExecuteSirenAction<LedgerAction>(
+      "deploy",
+      tryParams
+    );
+    await expect(executeActionPromise).resolves;
+
+    const promise = swap.deploy(tryParams);
+
+    await expect(promise).rejects.toBeInstanceOf(ChainError);
+
+    return { swapScope, actionScope };
   });
 });
--- a/src/swap.spec.ts
+++ b/src/swap.spec.ts
@@ -1,7 +1,7 @@
 import nock = require("nock");
 import { Problem } from "./cnd/axios_rfc7807_middleware";
-import { Cnd } from "./cnd/cnd";
-import { Swap } from "./swap";
+import { Cnd, SwapDetails } from "./cnd/cnd";
+import { ChainError, Swap } from "./swap";
 
 describe("Swap", () => {
   it("Throws a problem if cnd returns an error when executing an action", async () => {
@@ -34,6 +34,39 @@ describe("Swap", () => {
     });
 
     await expect(promise).rejects.toBeInstanceOf(Problem);
+
+    return scope;
+  });
+
+  it("Throws a ChainError if a wallet returns an error when executing an action", async () => {
+    const basePath = "http://example.com";
+    const cnd = new Cnd(basePath);
+
+    const selfPath = "/swap-id";
+
+    const swap = new Swap(cnd, `${basePath}/swap-id`, {});
+
+    const swapDetails: SwapDetails = {
+      actions: [
+        {
+          href: "/swap-id/action",
+          name: "action",
+          method: "GET",
+          fields: [{ name: "address", class: ["bitcoin", "address"] }]
+        }
+      ]
+    };
+
+    const scope = nock(basePath)
+      .get(selfPath)
+      .reply(200, JSON.stringify(swapDetails), {});
+
+    const promise = swap.tryExecuteSirenAction("action", {
+      maxTimeoutSecs: 1,
+      tryIntervalSecs: 0.01
+    });
+
+    await expect(promise).rejects.toBeInstanceOf(ChainError);
 
     return scope;
   });
--- a/src/swap.spec.ts
+++ b/src/swap.spec.ts
@@ -110,7 +110,7 @@ describe("Swap", () => {
       tryIntervalSecs: 0.01
     };
 
-    // To be sure we are testing the right thing, we verify this does not throws
+    // To be sure we are testing the right thing, we verify this does not throw
     const executeActionPromise = swap.tryExecuteSirenAction<LedgerAction>(
       "deploy",
       tryParams
--- a/src/swap.spec.ts
+++ b/src/swap.spec.ts
@@ -1,0 +1,40 @@
+import nock = require("nock");
+import { Problem } from "./cnd/axios_rfc7807_middleware";
+import { Cnd } from "./cnd/cnd";
+import { Swap } from "./swap";
+
+describe("Swap", () => {
+  it("Throws a problem if cnd returns an error when executing an action", async () => {
+    const basePath = "http://example.com";
+    const cnd = new Cnd(basePath);
+
+    const selfPath = "/swap-id";
+
+    const swap = new Swap(cnd, selfPath, {});
+
+    const scope = nock(basePath)
+      .get(selfPath)
+      .reply(
+        400,
+        JSON.stringify({
+          status: "400",
+          title: "Swap not supported",
+          type: "about:blank",
+          detail:
+            "The requested combination of ledgers and assets is not supported."
+        }),
+        {
+          "content-type": "application/problem+json"
+        }
+      );
+
+    const promise = swap.tryExecuteSirenAction("action", {
+      maxTimeoutSecs: 1,
+      tryIntervalSecs: 0.01
+    });
+
+    await expect(promise).rejects.toBeInstanceOf(Problem);
+
+    return scope;
+  });
+});
--- a/src/swap.spec.ts
+++ b/src/swap.spec.ts
@@ -1,7 +1,7 @@
 import nock = require("nock");
 import { Problem } from "./cnd/axios_rfc7807_middleware";
 import { Cnd, LedgerAction, SwapDetails } from "./cnd/cnd";
-import { ChainError, Swap } from "./swap";
+import { Swap, WalletError } from "./swap";
 
 describe("Swap", () => {
   it("Throws a problem if cnd returns an error when executing an action", async () => {
@@ -36,7 +36,7 @@ describe("Swap", () => {
     await expect(promise).rejects.toBeInstanceOf(Problem);
   });
 
-  it("Throws a ChainError if a wallet returns an error when resolving fields", async () => {
+  it("Throws a WalletError if a wallet returns an error when resolving fields", async () => {
     const basePath = "http://example.com";
     const cnd = new Cnd(basePath);
 
@@ -64,10 +64,10 @@ describe("Swap", () => {
       tryIntervalSecs: 0.01
     });
 
-    await expect(promise).rejects.toBeInstanceOf(ChainError);
+    await expect(promise).rejects.toBeInstanceOf(WalletError);
   });
 
-  it("Throws a ChainError if a wallet returns an error when executing an action", async () => {
+  it("Throws a WalletError if a wallet returns an error when executing an action", async () => {
     const basePath = "http://example.com";
     const cnd = new Cnd(basePath);
 
@@ -119,6 +119,6 @@ describe("Swap", () => {
 
     const promise = swap.deploy(tryParams);
 
-    await expect(promise).rejects.toBeInstanceOf(ChainError);
+    await expect(promise).rejects.toBeInstanceOf(WalletError);
   });
 });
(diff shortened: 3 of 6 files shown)


src/swap.ts
```diff
@@ -48,7 +55,7 @@ export class Swap {
    * This is only valid for ERC20 swaps.
    *
    * @param tryParams Controls at which stage the exception is thrown.
-   * @returns The hash of the transaction that was sent to the blockchain network.
+   * @returns The hash of the transaction that was sent to the blockchain network. Throws if ???
```
bonomat: Imho this can throw due to an error with cnd OR with the blockchain node . Same for the other LedgerActions

src/swap.ts
```diff
@@ -5,6 +5,11 @@ import { Field } from "./cnd/siren";
 import { sleep, timeoutPromise, TryParams } from "./util/timeout_promise";
 import { AllWallets, Wallets } from "./wallet";
 
+/**
+ * A helper type to return errors in case of failure with a wallet or blockchain action.
+ */
+export type Result = string | Error;
```
bonomat: Imho this is too rusty for JS. I think it's more common to just throw or return a tuple i.e. `(returnValue, error)` where both can be undefined. 

I'm happy to go with your proposal though if you feel this is the right way.

thomaseizinger:
Have a look at what I just wrote yesterday: https://github.com/comit-network/comit-js-sdk/blob/b25b53052a589d8cb71b7e2a48a4ebd6ea9ecaac/src/cnd/cnd.spec.ts#L72-L102

It is a setup function for a tests against the `Cnd` class that allow you to return whatever response you want over HTTP and check how `Cnd` reacts to that.

I feel like this is should allow you to write tests with a high level of confidence!

D4nte:
> Have a look at what I just wrote yesterday:
> 
> https://github.com/comit-network/comit-js-sdk/blob/b25b53052a589d8cb71b7e2a48a4ebd6ea9ecaac/src/cnd/cnd.spec.ts#L72-L102
> 
> It is a setup function for a tests against the `Cnd` class that allow you to return whatever response you want over HTTP and check how `Cnd` reacts to that.
> 
> I feel like this is should allow you to write tests with a high level of confidence!
> 
> You would probably want to generalize it by extracting out another function that allows you to call different methods on `Cnd` an not just `postSwap`.

Hi @thomaseizinger, unfortunately this does not help with the topics that are for review here. Please read the description, I am still deciding on the behaviour we are aiming for. I am well aware of your previous PR and familiar with `nock`. Thanks.

da-kami:
Can we also intercept all calls to the wallet in the way @thomaseizinger did it with the communication with cnd?

I feel, if we actively throw something it should be nicely wrapped into an Exception type (similar to what @thomaseizinger introduced). 

I hope I understood the problem correctly and my answer adds value :)

D4nte:
> Can we also intercept all calls to the wallet in the way @thomaseizinger did it with the communication with cnd?
>
> I feel, if we actively throw something it should be nicely wrapped into an Exception type (similar to what @thomaseizinger introduced).

You mean having something a bit more elaborate than the `Result` type I proposed that has an interface with various field. I can look into it but I am not sure what properties would make for such type.

> I hope I understood the problem correctly and my answer adds value :)

I feel like your answer does not really take in account the proposal here but just is what you had in mind for this problematic.

src/swap.ts
D4nte: I am sorry but this comment is confusing to me. I explicitly wrote in the description of the PR: "On functions that interacts with cnd and a ledger, cnd errors bubbles up via throws but ledger action are returned as result in case it returns an error."

So no, the aim would be for this function to return a `Promise<Result>` and not throw on blockchain errors, only throw on cnd errors.

src/swap.ts
D4nte: > Imho this is too rusty for JS. I think it's more common to just throw or return a tuple i.e. `(returnValue, error)` where either is defined.

Your tuple is just a different manner to define a data type `Result`.

src/swap.ts
bonomat: I wanted to finish your sentence as it seemed like you did not know what to write there.

thomaseizinger:
JS and TS don't have pattern matching to properly to conveniently unwrap something like string | Error.

I would stick to Exceptions, that is the go to error handling strategy in JS.

da-kami:
> > Can we also intercept all calls to the wallet in the way @thomaseizinger did it with the communication with cnd?
> > I feel, if we actively throw something it should be nicely wrapped into an Exception type (similar to what @thomaseizinger introduced).
> 
> You mean having something a bit more elaborate than the `Result` type I proposed that has an interface with various field. I can look into it but I am not sure what properties would make for such type.
> 
> > I hope I understood the problem correctly and my answer adds value :)
> 
> I feel like your answer does not really take in account the proposal here but just is what you had in mind for this problematic.

Concerning the conrete proposale, I think we have to think about what should be the outcome for the user:

Can the user actually act upon a specific ledger error's result? If yes then B, otherwise A. If there is nothing that the application can do then I'd rather go for unification (like @bonomat said). If there is some value for the application on top to get more specific information from a ledger's action failure then we should return it.

How do you imagine the ledger action failure result would look? Do you already have an example in mind?

src/swap.ts
D4nte: I see. The sentence is not finish because it is exactly the topic of this review. Do we throw everything (cnd and wallets errors), just cnd errors, nothing?

bonomat: Ah, got it. 
I would opt for throwing in all cases.

D4nte:
I will continue this work aiming to replicate @thomaseizinger PR #151 by extending the `Error` class to provide more contextual information and allow a bubble up (ie, throw everything).

yosriady:
I agree with @bonomat and @thomaseizinger that `export type Result = string | Error;` is a bit unwieldy because JS doesn't have type pattern matching. 

Promises are better represented with two type arguments. How about something like:

```
// if
fn1: () => Promise<B, Error1>
fn2: () => Promise<C, Error2>
// then
fn1(a).then(fn2): Promise<C, Error1 | Error2>

// if no errors are possible
fn3: () => Promise<D, never>
```

The Promise would return (`result`, `error`), similar to the 'nodebacks' which are idiomatic to the JS ecosystem. If there are no errors, `error` would be null.

I'm also in favour of bubbling up all errors for the SDK users to handle, because: As an app developer, I would need to display any errors to my users. If the code just throws any errors in-place, I have no way to do that. Even if there is nothing that the application can do, the SDK should still return some value for the application on top.

D4nte:
> I agree with @bonomat and @thomaseizinger that `export type Result = string | Error;` is a bit unwieldy because JS doesn't have type pattern matching.
> 
> Promises are better represented with two type arguments. How about something like:
> 
> ```js
> // if
> fn1: () => Promise<B, Error1>
> fn2: () => Promise<C, Error2>
> // then
> fn1(a).then(fn2): Promise<C, Error1 | Error2>
> 
> // if no errors are possible
> fn3: () => Promise<D, never>
> ```
> 
> The Promise would return (`result`, `error`), similar to the 'nodebacks' which are idiomatic to the JS ecosystem. If there are no errors, `error` would be null.
> 
> I'm also in favour of bubbling up all errors for the SDK users to handle, because: As an app developer, I would need to display any errors to my users. If the code just throws any errors in-place, I have no way to do that. Even if there is nothing that the application can do, the SDK should still return some value for the application on top.

Thank you so much @yosriady, this is great. However, I can't seem to be able to apply what you are proposing:

```
src/cnd/cnd.ts:268:29 - error TS2314: Generic type 'Promise<T>' requires 1 type argument(s).

268   public async getPeerId(): Promise<string, Problem | Error> {
                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Should I import a different `Promise` definition?

D4nte:
> > I agree with @bonomat and @thomaseizinger that `export type Result = string | Error;` is a bit unwieldy because JS doesn't have type pattern matching.
> > Promises are better represented with two type arguments. How about something like:
> > ```js
> > // if
> > fn1: () => Promise<B, Error1>
> > fn2: () => Promise<C, Error2>
> > // then
> > fn1(a).then(fn2): Promise<C, Error1 | Error2>
> > 
> > // if no errors are possible
> > fn3: () => Promise<D, never>
> > ```
> > 
> > 
> > The Promise would return (`result`, `error`), similar to the 'nodebacks' which are idiomatic to the JS ecosystem. If there are no errors, `error` would be null.
> > I'm also in favour of bubbling up all errors for the SDK users to handle, because: As an app developer, I would need to display any errors to my users. If the code just throws any errors in-place, I have no way to do that. Even if there is nothing that the application can do, the SDK should still return some value for the application on top.
> 
> Thank you so much @yosriady, this is great. However, I can't seem to be able to apply what you are proposing:
> 
> ```
> src/cnd/cnd.ts:268:29 - error TS2314: Generic type 'Promise<T>' requires 1 type argument(s).
> 
> 268   public async getPeerId(): Promise<string, Problem | Error> {
>                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> Should I import a different `Promise` definition?

Humm, not sure it's possible: https://github.com/Microsoft/TypeScript/issues/6283

D4nte:
> > > I agree with @bonomat and @thomaseizinger that `export type Result = string | Error;` is a bit unwieldy because JS doesn't have type pattern matching.
> > > Promises are better represented with two type arguments. How about something like:
> > > ```js
> > > // if
> > > fn1: () => Promise<B, Error1>
> > > fn2: () => Promise<C, Error2>
> > > // then
> > > fn1(a).then(fn2): Promise<C, Error1 | Error2>
> > > 
> > > // if no errors are possible
> > > fn3: () => Promise<D, never>
> > > ```
> > > 
> > > 
> > > The Promise would return (`result`, `error`), similar to the 'nodebacks' which are idiomatic to the JS ecosystem. If there are no errors, `error` would be null.
> > > I'm also in favour of bubbling up all errors for the SDK users to handle, because: As an app developer, I would need to display any errors to my users. If the code just throws any errors in-place, I have no way to do that. Even if there is nothing that the application can do, the SDK should still return some value for the application on top.
> > 
> > 
> > Thank you so much @yosriady, this is great. However, I can't seem to be able to apply what you are proposing:
> > ```
> > src/cnd/cnd.ts:268:29 - error TS2314: Generic type 'Promise<T>' requires 1 type argument(s).
> > 
> > 268   public async getPeerId(): Promise<string, Problem | Error> {
> >                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > ```
> > 
> > 
> > Should I import a different `Promise` definition?
> 
> Humm, not sure it's possible to specify the type of the rejection: [microsoft/TypeScript#6283](https://github.com/microsoft/TypeScript/issues/6283)

Ok looks like this works: 

```
  public async getPeerId(): Promise<string | Problem | Error> {
    const info = await this.getInfo();
    if (!info.id) {
      return Promise.reject(new Error("id field not present"));
    }

    return info.id;
  }
```
Thanks to https://www.stevefenton.co.uk/2017/11/getting-type-safety-for-typescript-promises/

yosriady:
@D4nte Could we use a [tuple type](https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple) instead? Like:

```
public async getPeerId(): Promise<[string, Error]>
```

src/cnd/cnd.ts
```diff
@@ -277,23 +289,53 @@ export class Cnd {
     return info.listen_addresses;
   }
 
+  /**
+   * Initiate a swap request on cnd.
```
thomaseizinger:
```suggestion
   * Initiate a swap on cnd.
```

Initiate and request have a similar meaning, we should use either one or the other.

Initiating a request is something different :)

src/cnd/axios_rfc7807_middleware.ts
```diff
@@ -15,6 +16,7 @@ export class Problem extends Error {
     this.detail = detail;
     this.instance = instance;
     this.title = title;
+    this.isProblem = true;
```
thomaseizinger: :mega: This seems to be better expressed at the type-level, which it is because it is a `Problem` type.

src/cnd/cnd.ts
```diff
@@ -259,15 +259,27 @@ export class Cnd {
...
+   * Get the peer id of the cnd node
+   *
+   * @returns Promise that resolves with the peer id of the cnd node,
+   * @throws A {@link Problem} from the cnd REST API or an Error.
+   */
   public async getPeerId(): Promise<string> {
     const info = await this.getInfo();
     if (!info.id) {
-      throw new Error("id field not present");
+      return Promise.reject(new Error("id field not present"));
```
bonomat: I did a quick google on this but didn't understand it :D 
Why are you using `Promise.reject` here and not `throw new Error`?

src/cnd/axios_rfc7807_middleware.ts
```diff
@@ -7,6 +7,7 @@ export class Problem extends Error {
   public readonly status?: number;
   public readonly detail?: string;
   public readonly instance?: string;
+  public readonly isProblem: boolean;
```
bonomat: From your PR description I take that you use this field to identify whether we have an `error` or a `problem`. If `Problem.isProblem` is always true we can assume it is always true and don't have to check for it right? Instead we can check if the `type of` the received exception is `Problem`. 

e.g. 

```
try {
	swap.redeem();
} catch (error) {
	if (error instanceof ChainError) {
		// do something
    } else if (error instanceof Problem) {
       // do something else
    } else {
      panicAtTheDiscotheque
	}
}
```

What do you think about this?

bonomat: Oh, @thomaseizinger had the same comment :)

src/cnd/cnd.ts
```diff
@@ -259,15 +259,27 @@ export class Cnd {
...
     }
 
     return info.id;
   }
 
+  /**
+   * Get the address on which cnd are listening for peer-to-peer/COMIT messages.
```
da-kami:
```suggestion
   * Get the address on which cnd is listening for peer-to-peer/COMIT messages.
```

src/cnd/cnd.ts
```diff
@@ -277,23 +289,53 @@ export class Cnd {
...
+   */
   public async getSwaps(): Promise<SwapSubEntity[]> {
     const response = await this.fetch("swaps");
     const entity = response.data as Entity;
 
     return entity.entities as SwapSubEntity[];
   }
 
+  /**
+   * Fetch data from the REST API
```
da-kami:
```suggestion
   * Fetch data from the REST API.
```

src/cnd/cnd.ts
```diff
@@ -277,23 +289,53 @@ export class Cnd {
...
+   * @throws Problem | Error A {@link Problem} from the cnd REST API or an Error.
+   */
   public fetch<T>(path: string): AxiosPromise<T> {
     return this.client.get(path);
   }
 
+  /**
+   * Proceed with an action request on the cnd REST API.
+   * This request could be a GET HTTP request, which only returns data needed to further proceed with the action,
+   * or a POST request, which is the action itself, to instruct cnd about a swap
```
da-kami:
```suggestion
   * or a POST request, which is the action itself, to instruct cnd about a swap.
```

src/cnd/cnd.ts
```diff
@@ -277,23 +289,53 @@ export class Cnd {
...
+   *
+   * @param action The action to perform
```
da-kami:
```suggestion
   * @param action The action to perform.
```

src/swap.ts
```diff
@@ -93,7 +115,9 @@ export class Swap {
    * If the {@link Swap} is not in the right state this call will throw a timeout exception.
    *
    * @param tryParams Controls at which stage the exception is thrown.
-   * @returns The hash of the transaction that was sent to the blockchain network.
+   * @returns The result to the refund action, either the hash of the transaction that was sent to the blockchain
```
da-kami:
```suggestion
   * @returns The result to the refund action, a hash of the transaction that was sent to the blockchain.
```

src/swap.ts
```diff
@@ -112,10 +142,15 @@ export class Swap {
...
    * If you are using any of the above actions ({@link Swap.redeem}, etc) you shouldn't need to use this.
+   * This only performs an action on the CND REST API, if an action is needed on another system (e.g blockchain wallet),
+   * then the needed information is returned by this function and needs to be processed with {@link doLedgerAction}.
    *
    * @param actionName The name of the Siren action you want to execute.
    * @param tryParams Controls at which stage the exception is thrown.
-   * @returns The response from {@link Cnd}. The actual response depends on the action you executed (hence the type parameter).
+   * @returns The response from {@link Cnd}. The actual response depends on the action you executed (hence the
+   * type parameter).
+   * @throws Error | Problem Throws if cnd returns an error or there is a network issue. If a POST is done the cnd API, it is
```
da-kami:
```suggestion
   * @throws Error | Problem Throws if cnd returns an error or there is a network issue. If a POST is done, it is
```

src/swap.ts
```diff
@@ -112,10 +142,15 @@ export class Swap {
...
+   * the last step and should be considered at failed if this throws.
```
da-kami:
```suggestion
   * the last step and should be considered as failed if this throws.
```

src/swap.ts
```diff
@@ -133,34 +168,59 @@ export class Swap {
    * Uses the wallets given in the constructor to send transactions according to the given ledger action.
    *
    * @param ledgerAction The ledger action returned from {@link Cnd}.
+   * @throws ChainError Throws if a wallet or blockchain action failed
```
da-kami:
```suggestion
   * @throws ChainError Throws if a wallet or blockchain action failed.
```

src/cnd/cnd.ts
D4nte: @da-kami Hum, why are you fixing typos at a 10% review? 🤔

src/cnd/cnd.ts
D4nte: It's an artefact from when I was playign with `reject`/`resolve` to try to implement Yos' proposal.

da-kami:
Actually: If this is still a 10% review request as suggested in the PR description I might have overshot with my comments... But from my understanding this is 99% now...

src/cnd/axios_rfc7807_middleware.ts
thomaseizinger: Stop reading my mind @bonomat, that is scary!

bonomat: Didn't have to. When reviewing code the voice in my head sounds like you 🙉🙊🙈

D4nte:
> Actually: If this is still a 10% review request as suggested in the PR description I might have overshot with my comments... But from my understanding this is 99% now...

The Description states _10%_ so not sure why you think it's 99?

src/swap.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    const promise = swap.tryExecuteSirenAction("action", {
+      maxTimeoutSecs: 1,
+      tryIntervalSecs: 0.01
+    });
+
+    try {
+      await promise;
+    } catch (e) {
+      console.log(JSON.stringify(e, undefined, 2));
```
D4nte: @thomaseizinger this `e` is not a `Problem` but an `Error`. I have done another tests that check the instance of it and it fails too. The following is printed here:

```
 {
      "message": "Request failed with status code 400",
      "name": "Error",
      "stack": "Error: Request failed with status code 400\n    at createError (/Users/froyer/src/comit-network/comit-js-sdk/node_modules/axios/lib/core/createError.js:16:15)\n    at settle (/Users/froyer/src/comit-network/comit-js-sdk/node_modules/axios/lib/core/settle.js:17:12)\n    at IncomingMessage.handleStreamEnd (/Users/froyer/src/comit-network/comit-js-sdk/node_modules/axios/lib/adapters/http.js:236:11)\n    at IncomingMessage.emit (events.js:311:20)\n    at endReadableNT (_stream_readable.js:1204:12)\n    at processTicksAndRejections (internal/process/task_queues.js:84:21)",
      "config": {
        "url": "/swap-id",
        "method": "get",
        "headers": {
          "Accept": "application/json, text/plain, */*",
          "User-Agent": "axios/0.19.2"
        },
        "baseURL": "http://example.com",
        "transformRequest": [
          null
        ],
        "transformResponse": [
          null
        ],
        "timeout": 0,
        "xsrfCookieName": "XSRF-TOKEN",
        "xsrfHeaderName": "X-XSRF-TOKEN",
        "maxContentLength": -1
      }
    }
```

I was expecting a problem to be thrown by `Cnd.fetch`. What am I missing?

D4nte: Ignore me!

src/swap.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+      );
+
+    const promise = swap.tryExecuteSirenAction("action", {
+      maxTimeoutSecs: 1,
+      tryIntervalSecs: 0.01
+    });
+
+    await expect(promise).rejects.toBeInstanceOf(Problem);
+
+    return scope;
```
bonomat: Is there a particular reason why you return `scope` here? 
I quickly looked it [up](https://github.com/nock/nock#isdone): you can call `scope.isDone()` to ensure that the nock expectation was met.

src/swap.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+      .get("/swap-id/deploy")
+      .times(2)
+      .reply(200, JSON.stringify(actionPayload), {});
+
+    const tryParams = {
+      maxTimeoutSecs: 1,
+      tryIntervalSecs: 0.01
+    };
+
+    // To be sure we are testing the right thing, we verify this does not throws
```
bonomat:
```suggestion
    // To be sure we are testing the right thing, we verify this does not throw
```

src/swap.ts
```diff
@@ -103,6 +119,12 @@ export class Swap {
     return this.doLedgerAction(response.data);
   }
 
+  /**
+   * Fetch the details of a swap.
+   *
+   * @return The details of the swap at returned by cnd REST API.
```
bonomat:
```suggestion
   * @return The details of the swap are returned by cnd REST API.
```

src/cnd/cnd.ts
```diff
@@ -277,23 +289,53 @@ export class Cnd {
...
+   * @returns The data returned by cnd.
+   * @throws A {@link Problem} from the cnd REST API or an {@link Error}.
+   */
   public fetch<T>(path: string): AxiosPromise<T> {
     return this.client.get(path);
   }
 
+  /**
+   * Proceed with an action request on the cnd REST API.
+   * This request could be a GET HTTP request, which only returns data needed to further proceed with the action,
```
thomaseizinger: This documentation knows more than the function itself (who says we don't do PUT/DELETE requests at some point?). I would discourage from doing that because it makes maintaining the documentation very hard in the long run.

I vaguely remember @da-kami and me actually documenting that function, where did that go?

src/swap.ts
```diff
@@ -5,6 +5,16 @@ import { Field } from "./cnd/siren";
 import { sleep, timeoutPromise, TryParams } from "./util/timeout_promise";
 import { AllWallets, Wallets } from "./wallet";
 
+export class ChainError extends Error {
+  constructor(
+    public readonly attemptedAction: string,
+    public readonly callError: Error,
+    public readonly callParams: any
+  ) {
+    super(callError.message);
```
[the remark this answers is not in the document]
D4nte: According to your previous comment, you propose `source` to be the `Error`.

`new Error()` takes one `string` argument so no, I cannot call `super(source)`.

src/swap.ts
```diff
@@ -255,9 +339,15 @@ export class Swap {
...
       }
 
-      return this.cnd.executeSirenAction(action!, (field: Field) =>
-        this.fieldValueResolver(field)
-      );
+      // This throws if cnd returns an error or there is a network error
+      return this.cnd.executeSirenAction(action!, async (field: Field) => {
+        try {
+          // Awaiting here allows us to have better context
+          return await this.fieldValueResolver(field);
```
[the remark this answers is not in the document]
D4nte: Potato potato, I had both ways locally, I don't see a difference except for my idea saying `Local variable 'value' is redundant` with your proposal.

src/cnd/cnd.ts
D4nte: > I vaguely remember @da-kami and me actually documenting that function, where did that go?

I am not sure what you want me to say. You can see from this diff that there were no existing comment.

src/swap.ts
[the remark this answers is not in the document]
thomaseizinger: Fair enough. Does TSLint not complain about `return await`?

src/swap.ts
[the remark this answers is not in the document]
thomaseizinger: Sometimes, JS really sucks :sweat_smile:

src/swap.ts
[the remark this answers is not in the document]
D4nte: No issue or warning have been returned by our tools for the `return await`.

src/swap.spec.ts
```diff
@@ -0,0 +1,130 @@
+import nock = require("nock");
+import { Problem } from "./cnd/axios_rfc7807_middleware";
+import { Cnd, LedgerAction, SwapDetails } from "./cnd/cnd";
+import { ChainError, Swap } from "./swap";
+
+describe("Swap", () => {
```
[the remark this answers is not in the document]
D4nte: Now sorted, please review.

src/swap.ts
```diff
@@ -5,6 +5,16 @@ import { Field } from "./cnd/siren";
 import { sleep, timeoutPromise, TryParams } from "./util/timeout_promise";
 import { AllWallets, Wallets } from "./wallet";
 
+export class WalletError extends Error {
+  constructor(
+    public readonly attemptedAction: string,
+    public readonly source: Error,
+    public readonly callParams: any
+  ) {
+    super(JSON.stringify(source));
```
thomaseizinger: :mega: Sorry for sending you on a wrong path here with the source. If JS Errors don't support being chained, we should just pass our own message and be done but definitely not JSON.stringifing the error type. That can lead to runtime errors if the type has self-references.

src/swap.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const promise = swap.tryExecuteSirenAction("action", {
+      maxTimeoutSecs: 1,
+      tryIntervalSecs: 0.01
+    });
+
+    await expect(promise).rejects.toBeInstanceOf(WalletError);
+  });
+
+  it("Throws a WalletError if a wallet returns an error when executing an action", async () => {
```
thomaseizinger: Which code path we trying to test here? This one? https://github.com/comit-network/comit-js-sdk/pull/152/files#diff-7ca577257aac4fa5e3c549c84277afc6R210-R218

Just make the EthereumWallet throw and exception, why tinker around with `Cnd`?

src/swap.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const promise = swap.tryExecuteSirenAction("action", {
+      maxTimeoutSecs: 1,
+      tryIntervalSecs: 0.01
+    });
+
+    await expect(promise).rejects.toBeInstanceOf(Problem);
+  });
+
+  it("Throws a WalletError if a wallet returns an error when resolving fields", async () => {
```
thomaseizinger: You could just mock the `fieldValueResolver` instead of `Cnd`.

src/swap.spec.ts
D4nte: Unfortunately your link does not work for me. Are you talking about testing `doLedgerAction` directly?

src/swap.spec.ts
D4nte: The first step of `Swap.tryExecuteSirenAction` is `Swap.fetchDetails` so I do need to mock `Cnd` first. I am not sure to understand how you could test this without mocking `Cnd`.

src/swap.spec.ts
thomaseizinger: I am talking about making the actual wallet action throw an exception and assert that is correctly wrapped in a `WalletError`.

D4nte: It does throw an exception as it is not initialised but I understand your point as it is not explicit in the test.

Issue #149, opened 2020-03-04
Properly use bcoin library


D4nte:
## Problem

The bitcoin wallet we offer does not properly use the bcoin library. Indeed, it was pointed to use that we should use the `SPVNode` class but we do not.

To run this on testnet, we actually had to use the `SPVNode` class: https://github.com/comit-network/testnet-demo/blob/8b925ed0d64637bb32564d25d844196afae9ecf4/src/bcoinWallet.ts#L30, this should be enough to make us think that maybe the currently offered wallet is not up to scratch.

## Goal

The SDK should be testnet ready and the difference of BitcoinWallet implementation between this and testnet-demo folder clearly show that there are potential issues.

## Recommendation

Bring the code over from testnet-demo.

D4nte:
@bonomat I propose to close this with https://github.com/comit-network/comit-js-sdk/pull/153 to highlight the fact that we do not expect someone to use this Bitcoin Wallet implementation for mainnet purposes.

bonomat:
Agreed.

PR #153, opened 2020-03-09
Better documentation of provided BitcoinWallet

D4nte:
Resolves https://github.com/comit-network/comit-js-sdk/issues/149


yosriady:
Should we mention this requirement as part of an app tutorial?

D4nte:
> Should we mention this requirement as part of an app tutorial?

Yes definitely

PR #158, opened 2020-03-18
Define payload for ln endpoints

D4nte:
Resolves comit-network/comit-rs#2154.

Please review, it's a draft to be sure it does not get merged with the comments.

If a field is commented it just means that I considered it and decided against including it.

I basically look at the existing swap request and the HALight action payloads, copied all the information in and reviewed each field.


src/cnd/cnd.ts
```diff
@@ -44,6 +44,50 @@ export interface SwapRequest {
...
 
+export interface CoreRequestBody {
+  // to_public_key: string; // This is exchanged via libp2p
+  // secret_hash: string; // Generated by cnd
+  // alpha_ledger: string; // Ledger name is in the route
+  // beta_ledger: string; // Ledger name is in the route
+  alpha_amount: string; // Asset name is in the route
+  beta_amount: string; // Asset name is in the route
+  // final_cltv_delta: number; // beta_expiry
+  // cltv_expiry: number; // beta_expiry
```
thomaseizinger: Two times beta_expiry here?

src/cnd/cnd.ts
```diff
@@ -44,6 +44,50 @@ export interface SwapRequest {
   peer: Peer;
 }
 
+export interface CoreRequestBody {
+  // to_public_key: string; // This is exchanged via libp2p
+  // secret_hash: string; // Generated by cnd
+  // alpha_ledger: string; // Ledger name is in the route
+  // beta_ledger: string; // Ledger name is in the route
+  alpha_amount: string; // Asset name is in the route
```
thomaseizinger: At the expense of having to define more interfaces: should we consider _not_ having alpha/beta terminology in the API but just say:

`satoshis` for example?

Or maybe a combination of both worlds? Although that would be quite weird.

The motivation would be that we don't need to try and maintain a terminology across all ledgers and protocols and fit it in. A good example is `final_cltv_expiry`, how is the user going to know what should go in this field?

Maybe the current approach is fine, I just wanted to highlight that with different routes per combination, we have the freedom to do something different.

src/cnd/cnd.ts
D4nte: It's because initially I was only building the payload for `HanEthereumEtherHalightLightningBitcoin`.

At the end, this is the `alpha/beta_expiry` so we do not need to include it in the payload. You did notice this is commented right?

src/cnd/cnd.ts
D4nte: > At the expense of having to define more interfaces: should we consider _not_ having alpha/beta terminology in the API but just say:
> 
> `satoshis` for example?
> 
> Or maybe a combination of both worlds? Although that would be quite weird.
> 
> The motivation would be that we don't need to try and maintain a terminology across all ledgers and protocols and fit it in. A good example is `final_cltv_expiry`, how is the user going to know what should go in this field?

Hum, sorry, what `final_cltv_expiry` field? There is no such field defined in this proposal.

> 
> Maybe the current approach is fine, I just wanted to highlight that with different routes per combination, we have the freedom to do something different.

True. I went for homogeneity across the route but it may not be justifiable.
The issues with `satoshis` is: What would we propose for ERC-20 tokens?
What if we are dealing with WBTC? it might create some confusions I belive.

src/cnd/cnd.ts
thomaseizinger: I was referring to `final_cltv_delta` and `cltv_expiry` both saying "beta_expiry". I assumed this meant that the values of these fields go into "beta_expiry" but we need both for LN so how is this going to work?

D4nte: Ah, ok I think I start to understand the issue.

`final_cltv_delta` and `cltv_expiry` are the same things and must have the same value so if we are doing HanHALight then `beta_expiry` holds the value for both (recip. HALightHAn, `alpha_expiry` holds the value).

src/cnd/cnd.ts
thomaseizinger: > What would we propose for ERC-20 tokens? What if we are dealing with WBTC? it might create some confusions I believe.

Very good questions!
How much time do you recon is it worth to invest now here instead of making breaking changes later?

D4nte: > > What would we propose for ERC-20 tokens? What if we are dealing with WBTC? it might create some confusions I believe.
> 
> Very good questions!
> How much time do you recon is it worth to invest now here instead of making breaking changes later?

Well to avoid breaking changes I'd be keen to use `alpha_amount` terminology.

thomaseizinger: Okay, let's roll with it and see where we end up :)

src/cnd/cnd.ts
thomaseizinger: I am indecisive here. Should we name it `beta_cltv_expiry` then? I feel like there should be a strong link in the name to what you should put here.

D4nte: It's a question of how precise you want the name to be on a given route vs homogeneity across the routes.
I don't know, maybe one of the other reviewers wants to chime in?

thomaseizinger: @bonomat @luckysori @da-kami !

da-kami: I have to admit that I don't understand why you would name it `beta_cltv_expiry` instead of `cltv_expiry` @thomaseizinger . From @D4nte 's comment I would have understood that it could be for both alpha or beta.
I might need more explanation for this to understand the problem so I can give meaningful advice.

thomaseizinger: It would be prefixed with alpha or beta depending in which ledger Lightning is!

That is kind of the combination of:
- use domain specific wording to be clearer
- use alpha/beta to be unambigous

da-kami: OK, I get it. That would mean the `RequestBody` would be different depending on if Lightning is alpha or beta. 

One follow up question:
Is the defined request body only for cases where Lightning is beta ledger? If so I think we should make it specific like suggested by @thomaseizinger

D4nte: It's a currently defined across but we can change.

src/cnd/cnd.ts
```diff
@@ -44,6 +44,50 @@ export interface SwapRequest {
...
+  // Instead of making it optional for all route it should be mandatory
+  // for the route that uses it.
+  alpha_contract_address: string;
+}
+
+export interface BetaHerc20AdditionalRequestBody {
+  beta_contract_address: string;
+}
+
+export type HanEthereumEtherHalightLightningBitcoinRequestBody = CoreRequestBody;
```
luckysori: :question: Isn't it more like this?

```suggestion
export type HanEthereumEtherHalightRequestBody = CoreRequestBody;
```

The protocol implies the ledger and the asset in the case of HALight.

src/cnd/cnd.ts
D4nte: Can someone remind me what expiry we use for Bitcoin and Ethereum? Absolute or relative?

luckysori: I would also advocate for no alpha or beta terminology and just have one interface per protocol combination.

src/cnd/cnd.ts
```diff
@@ -44,6 +44,50 @@ export interface SwapRequest {
...
+  alpha_expiry: number; // Fits in JS numbers
+  beta_expiry: number; // Fits in JS numbers
+  // chain: string; // (from lnd action payload) In the URL
+  alpha_network: string; // For Ethereum we use "chain" terminology but we could revert to "network" on this new route to keep same terminology across all ledgers
+  beta_network: string; // The bitcoin network
+  // Identities are always mandatory even if cnd could discover it (for lnd).
+  // So that the interface is simpler (no optional fields) and also adds an element of security (ensure it's connected to the right lnd).
+  // Also we dropped the `redeem` and `refund` terminology as they are redundant with the `alpha/beta` qualifier.
+  // Finally, this interface is to be used by both Bob and Alice so using the `refund`/`redeem` terminology would just make it more complex
+  alpha_ledger_identity: string;
```
luckysori: :question: If we decide to keep alpha, how about this?

```suggestion
  alpha_identity: string;
```

src/cnd/cnd.ts
luckysori: > Can someone remind me what expiry we use for Bitcoin and Ethereum? Absolute or relative?

Absolute timestamps.

D4nte: > I would also advocate for no alpha or beta terminology and just have one interface per protocol combination.

I don't think your proposal is feasible. Please provide example on what the fields would be for han/han combination.

D4nte: Please see 60ed7015 for a new proposal.

src/cnd/cnd.ts
D4nte: We could support Litecoin over Lightning
Can we support raiden over Halight?

D4nte:
Removing comments in 8bab602038c83bb7d638b3c3130a7a5df5e9eb61 to mark PR as ready for review to hopefully get more decisive reviews.

src/cnd/cnd.ts
luckysori: >  Please provide example on what the expiry fields would be for han/han combination

For HAN bitcoin: `bitcoin_expiry`.
For HAN ether: `ethereum_expiry`.

src/cnd/cnd.ts
luckysori: I guess being specific can't hurt.

src/cnd/cnd.ts
D4nte: > > Please provide example on what the expiry fields would be for han/han combination
> 
> For HAN bitcoin: `bitcoin_expiry`.
> For HAN ether: `ethereum_expiry`.

And for halight?

thomaseizinger: > For HAN bitcoin: bitcoin_expiry.
> For HAN ether: ethereum_expiry.

The downside of this naming-scheme is that it upfront prevents any combinations that act on the same ledger (btc<->ln for example). But maybe that limitation is ok, we probably cannot cater for everything.

luckysori:
I understand this was part of the original issue:

> Aim for a flat JSON design (i.e. no nested objects).

But what is the harm in going for something like this:

```ts
interface HanEthereumEtherHalightLightningBitcoin {
    alpha: HanEthereumEther
    beta: HalightLightningBitcoin
    role: "Alice" | "Bob";
    peer: Peer;
}

interface HanEthereumEther {
    amount: string;
    identity: string;
    expiry: number;
    chain: string;
}

interface HalightLightningBitcoin {
    amount: string;
    network: string;
    identity: string;
    cltv_expiry:
}
```

da-kami:
> I understand this was part of the original issue:
> 
> > Aim for a flat JSON design (i.e. no nested objects).
> 
> But what is the harm in going for something like this:
> 
> ```ts
> interface HanEthereumEtherHalightLightningBitcoin {
>     alpha: HanEthereumEther
>     beta: HalightLightningBitcoin
>     role: "Alice" | "Bob";
>     peer: Peer;
> }
> 
> interface HanEthereumEther {
>     amount: string;
>     identity: string;
>     expiry: number;
>     chain: string;
> }
> 
> interface HalightLightningBitcoin {
>     amount: string;
>     network: string;
>     identity: string;
>     cltv_expiry: number;
> }
> ```

I like this schema better than the one proposed in the PR. It focusses more on the protocols and is easier to understand. Why was the aim to keep the Json flat @thomaseizinger ?

D4nte:
> I understand this was part of the original issue:
> 
> > Aim for a flat JSON design (i.e. no nested objects).
> 
> But what is the harm in going for something like this:
> 
> ```ts
> interface HanEthereumEtherHalightLightningBitcoin {
>     alpha: HanEthereumEther
>     beta: HalightLightningBitcoin
>     role: "Alice" | "Bob";
>     peer: Peer;
> }
> 
> interface HanEthereumEther {
>     amount: string;
>     identity: string;
>     expiry: number;
>     chain: string;
> }
> 
> interface HalightLightningBitcoin {
>     amount: string;
>     network: string;
>     identity: string;
>     cltv_expiry:
> }
> ```

I think I turned around on the use of `bitcoin` instead of `alpha`. Alpha/Beta are concepts that we have introduced and one of the first hurdle to go for someone to understand COMIT.

Removing such terminology from the body is attractive as it is one less things from someone to learn to do their first swap (if we exclude the SDK as a thought experiment). Knowing that in any case, the URL is still in the `Alpha/Beta` form.

This makes me wonder. Should we keep the URL in the `Alpha/Beta` form? Now that both parties will have to send a similar message to their cnd. This message containing the role of course. Then I wonder whether our us of `Alpha/Beta` form instead of `Sell/Buy` or `Send/Receive` really makes sense.


It could be possible to say that the format is `Buy/Sell` so that If Alice sells LNBTC to Bob to buy PAY from him.

Alice sends to her cnd:
```
POST /herc20/ethereum/erc20/halight/bitcoin/bitcoin
{
	role: "Alice",
	bitcoin: "100000000",
	bitcoin_expiry: 3600,
	ethereum_erc20_token: "9000000000000000000000",
	ethereum_erc20_contract: "0x...",
	ethereum_expiry: 7200,
	peer: { bob_details }
}
```

Bob sends to his cnd:
```
POST /halight/bitcoin/bitcoin/herc20/ethereum/erc20
{
	role: "Bob",
	bitcoin: "100000000",
	bitcoin_expiry: 3600,
	ethereum_erc20_token: "9000000000000000000000",
	ethereum_erc20_contract: "0x...",
	ethereum_expiry: 7200,
	peer: { alice_details }
}
```

thomaseizinger:
> I like this schema better than the one proposed in the PR. It focusses more on the protocols and is easier to understand. Why was the aim to keep the Json flat @thomaseizinger ?

@da-kami A flat JSON is easier to reason about _if_ it is conceptually flat, i.e. the fields are actually orthogonal to each other. If we just force a flat JSON by encoding more information in a single field name, then there is not much won I think :)

Another reason for a flat JSON was to allow those endpoints to be described with Siren actions. That would allow us to build the SDK in a way so it doesn't have to know about those routes and can just discover through available actions, which swaps it can create and automatically fill in fields like identities for the user.

That being said, if the best model is not a flat JSON, we can definitely go for a nested design.

thomaseizinger:
> I understand this was part of the original issue:
> 
> > Aim for a flat JSON design (i.e. no nested objects).
> 
> But what is the harm in going for something like this:
> 
> ```ts
> interface HanEthereumEtherHalightLightningBitcoin {
>     alpha: HanEthereumEther
>     beta: HalightLightningBitcoin
>     role: "Alice" | "Bob";
>     peer: Peer;
> }
> 
> interface HanEthereumEther {
>     amount: string;
>     identity: string;
>     expiry: number;
>     chain: string;
> }
> 
> interface HalightLightningBitcoin {
>     amount: string;
>     network: string;
>     identity: string;
>     cltv_expiry: number;
> }
> ```

@luckysori No real harm, the question is: what is the benefit of a consistent structure (alpha/beta nested objects) throughout all routes if they don't have to be consistent?

thomaseizinger:
> Removing such terminology from the request body is attractive as it is one less thing to learn before doing a first swap (if we exclude the SDK as a thought experiment), because the URL is still in the Alpha/Beta form.

> This makes me wonder. Should we keep the URL in the Alpha/Beta form?

Generally, I think we should aim for users to have to care _less_ about the URL structure and not more.

I would prefer to stay in the domain of swaps (i.e. alpha/beta) for now. Once we add orders to COMIT, we can introduce a buy/sell terminology.

D4nte:
_Not quoting but basically continuing the discussion above_

If we want to be able to build siren action to discover the struct then I guess the naming of the field would not matter too much as we would use class attributes on the fields (in the siren payload) to convey their meaning ("bitcoin-amount", "satoshi-unit", etc). It might actually make it easier of we use `alpha`/`beta` terminology.

I think it's quite hard to predict here, is `bitcoin_expiry` (etc) better than `alpha_expiry`?

Maybe let's move away from the "brand new" API context and refer back to the existing APIs. We currently use `alpha`/`beta` in our payload. Has this led to any issue? If no, then let's keep it this way and not move to `bitcoin_` and create new potential issues.
If yes, then let's review the issues it created and find a better way.

D4nte:
@thomaseizinger @da-kami @luckysori 

We have made a number of considerations and discussed them.
It looks like to me the current design fit the purpose. We have discussed alternatives but none of them are significantly better than the current design.

If you think that I missed something and that part of the design should be further discussed, please raise your voice, otherwise, please approve.

luckysori:
> > I understand this was part of the original issue:
> > > Aim for a flat JSON design (i.e. no nested objects).
> > 
> > 
> > But what is the harm in going for something like this:
> > ```ts
> > interface HanEthereumEtherHalightLightningBitcoin {
> >     alpha: HanEthereumEther
> >     beta: HalightLightningBitcoin
> >     role: "Alice" | "Bob";
> >     peer: Peer;
> > }
> > 
> > interface HanEthereumEther {
> >     amount: string;
> >     identity: string;
> >     expiry: number;
> >     chain: string;
> > }
> > 
> > interface HalightLightningBitcoin {
> >     amount: string;
> >     network: string;
> >     identity: string;
> >     cltv_expiry: number;
> > }
> > ```
> 
> @luckysori No real harm, the question is: what is the benefit of a consistent structure (alpha/beta nested objects) throughout all routes if they don't have to be consistent?

Maybe I'm misunderstanding your point, but whether you nest it or not I do see consistent structure for all swap pairs: info related to one ledger; info related to other ledger; and the rest.

The argument for a nested design like the one I proposed for me would be composability.

src/cnd/cnd.ts
```diff
@@ -44,6 +44,56 @@ export interface SwapRequest {
   peer: Peer;
 }
 
+export interface CoreRequestBody {
+  alpha_amount: string;
+  beta_amount: string;
+  alpha_network: string;
+  beta_network: string;
+  alpha_identity: string;
```
thomaseizinger: Hmm, I just realized that this cannot be called `redeem`/`refund` because that is different for Alice and Bob right?

I guess it doesn't really matter, what we are using the address for right?

src/cnd/cnd.ts
```diff
@@ -44,6 +44,56 @@ export interface SwapRequest {
...
+export interface AlphaHanAdditionalRequestBody {
+  alpha_absolute_expiry: number;
+}
+
+export interface BetaHanAdditionalRequestBody {
+  beta_absolute_expiry: number;
+}
+
+export interface AlphaHerc20AdditionalRequestBody {
+  alpha_contract_address: string;
```
thomaseizinger: Make this more obvious?

```suggestion
  alpha_erc20_contract_address: string;
```

src/cnd/cnd.ts
D4nte: Yes correct, I do not use the `redeem`/`refund` because it means that you would expect different type of identity for Alice and Bob for the same swap. Also, the information  is redundant if you know the role and alpha/beta.

src/cnd/cnd.ts
D4nte: Yes I hesitated. Went for the simpler version for 2 reasons:
- Simpler is better
- The information that it is erc20 is already present in the URL

src/cnd/cnd.ts
thomaseizinger: I was more thinking from the user's perspective: The question we have to answer is "does the user care, for what the identity is used". If no, then this is fine, if yes we should change it to make it obvious.

The client shouldn't derive things from the URL IMO.

PR #164, opened 2020-03-23
Return new `Transaction` class from Ethereum actions

### LINKED ISSUE #107: Provide means to check for Ethereum transaction status
## Problem

When sending transactions, the internal `EthereumWallet` does not check for the transaction status. Meaning that if the rpc call is successful but the transaction fails, it goes unnoticed.

This is an issue that has been the cause of time wasting in e2e tests. The deployed failed but was marked as successful, making redeem/refund failed.
Investigation is then misdirected to the redeem/refund section of the deploy one.

## Goal

Report if the Ethereum transaction fails or give the lib consumer the opportunity to check themselves.

## Recommendation

```
const fundingTransaction =  await swap.fund();
> console.log(await fundingTransaction.getStatus())
"transaction failed" / "transaction confirmed" /  "transaction unconfirmed"
```

Which means a new `SwapTransaction` class could be introduced.

This change should be tested within the SDK.

Note: This is a recommendation. The objective is to provide an interface to get the status of a transaction without asking the app dev to do a lot of manual process, ie, returning the txid is not good enough.
Note: To test we need to introduce testcontainers js to test against an Ethereum node. Inspire yourself from bobtimus project.
Note: Don't worry about providing api for number of confirmations for now.
Note: This is only for Ethereum
Note: a transaction can become unconfirmed so always check the blockchain.

D4nte:
Resolves #107.

### DIFF
--- a/src/swap.ts
+++ b/src/swap.ts
@@ -1,6 +1,6 @@
 import { AxiosResponse } from "axios";
 import { BigNumber } from "bignumber.js";
-import { Cnd, LedgerAction, SwapDetails } from "./cnd/cnd";
+import { Cnd, LedgerAction, ledgerIsEthereum, SwapDetails } from "./cnd/cnd";
 import { Field } from "./cnd/siren";
 import Transaction from "./transaction";
 import { sleep, timeoutPromise, TryParams } from "./util/timeout_promise";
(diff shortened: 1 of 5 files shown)


src/swap.ts
```diff
@@ -322,6 +337,123 @@ export class Swap {
     }
   }
 
+  /**
+   * Get the Alpha deploy transaction.
+   *
+   * @returns null if cnd hasn't seen a deploy transaction, otherwise, {@link Transaction} if supported or the transaction id as string.
+   */
+  public async getAlphaDeployTransaction(): Promise<
```
thomaseizinger: At some point we will have to think how we are going to `scale` the `Swap` class with more protocols being added and all of them having different actions ..

D4nte: 1. I hesitated between fully-qualified function names or passing a `Alpha`/`Beta` and `Deploy`/`Fund`/`Redeem`/`Refund`.
2. I would like to see the actions going away once @yosriady is done with the first PoC and is keen to give an idea of the ideal interface.

D4nte:
bors

PR #171, opened 2020-03-25
Fix rejection on getting Ethereum transaction status

D4nte:
Ready for review, usage is in https://github.com/comit-network/comit-rs/pull/2313

A supplementary API could be make `Transaction` an event emitter. At this stage I'd prefer to see the need for it from the PoC before investigating time in it.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -6,13 +6,17 @@ export enum TransactionStatus {
    */
   Failed,
   /**
-   * The transaction was accepted by the blockchain node but yet to be mined.
+   * The transaction was not yet mined and its status is uncertain.
    */
   Pending,
   /**
    * The transaction was mined.
    */
-  Confirmed
+  Confirmed,
+  /**
+   * The transaction could not be retrieved.
+   */
+  NotFound
 }
 
 /**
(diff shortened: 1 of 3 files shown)


D4nte:
WIP, will talk with @yosriady to agree on best design.

src/transaction.ts
```diff
@@ -34,27 +34,56 @@ export class Transaction {
...
   }
 
+  private async ethereumWait(): Promise<void> {
+    const wallet = this.ethereumWallet;
+
+    const transaction = await wallet.getTransaction(this.id);
+    try {
+      await transaction.wait(1);
+    } catch (e) {
+      // If it rejects it means it's mined and failed.
```
da-kami: ❓ Are we 100% sure we want to swallow that exception? Shouldn't there be at least a log statement?

D4nte: The returned status will tell you if it is failed so it's not "swallowed".

Sorry I should have removed review request, redesigning with @yosriady's input.

src/transaction.ts
```diff
@@ -4,15 +4,19 @@ export enum TransactionStatus {
   /**
    * The transaction was rejected by the blockchain node.
    */
-  Failed,
+  Failed = "failed",
```
thomaseizinger: This is s a breaking change in TypeScript AFAIK.

thomaseizinger:
The "request changes" are concerning the breaking change which should be mentioned in the changelog.

src/transaction.ts
D4nte: Indeed thanks, I'll revert. I changed for a idea that I did not implement.

src/transaction.ts
```diff
@@ -25,34 +29,56 @@ export class Transaction {
...
+        // If the transaction is not yet mined then its status is uncertain.
+        return TransactionStatus.Pending;
+      }
+      return checkReceipt(this.id);
+    } else {
+      try {
+        await transaction.wait(confirmations);
+        return checkReceipt(this.id);
+      } catch (e) {
+        // Throws if the transaction is failed
```
thomaseizinger: The comment is lying :)

PR #176, opened 2020-03-28
Add close method for bcoin wallet

luckysori:
These are a couple of patches that need to be released before comit-network/comit-rs#2330 can be merged in.

### DIFF
--- a/src/wallet/bitcoin.ts
+++ b/src/wallet/bitcoin.ts
@@ -110,11 +110,9 @@ export class InMemoryBitcoinWallet implements BitcoinWallet {
   private constructor(
     public readonly network: any,
 
-    // @ts-ignore
     private readonly walletdb: any,
     private readonly pool: any,
 
-    // @ts-ignore
     private readonly chain: any,
     private readonly wallet: any
   ) {}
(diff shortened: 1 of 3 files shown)


src/wallet/bitcoin.ts
```diff
@@ -170,6 +168,23 @@ export class InMemoryBitcoinWallet implements BitcoinWallet {
     return "150";
   }
 
+  public async close(): Promise<any> {
+    const tasks = [];
+    if (this.pool.opened) {
+      tasks.push(await this.pool.close());
```
bonomat: I don't think that is what you want. You want to collect all the pending promises and `await` them all together at the end. 
I.e. 
```suggestion
      tasks.push(this.pool.close());
```

src/wallet/bitcoin.ts
```diff
@@ -170,6 +168,23 @@ export class InMemoryBitcoinWallet implements BitcoinWallet {
...
+
+    if (this.chain.opened) {
+      tasks.push(await this.chain.close());
+    }
+
+    if (this.walletdb.opened) {
+      tasks.push(await this.walletdb.close());
+    }
+
+    return Promise.all(tasks);
```
bonomat: The return value is not needed (we probably don't care anyways as we want to close everything). 
```suggestion
    await Promise.all(tasks);
```

if you would be doing 
```
return Promise.all(tasks);
```
the return value would be an array of all return values from the resolved promises above.

src/wallet/bitcoin.ts
luckysori: You're totally right. Exactly what I wanted. Bloody JS.

src/wallet/bitcoin.ts
luckysori: Not sure about this one. It ultimately doesn't matter, but all the other async functions in this file return a promise, rather than awaiting at the end. For consistency I would prefer to keep it as it is.

With your change I believe JS awaits inside the function and then wraps the result inside a promise that will resolve immediately and returns this value. That feels worse to me.

luckysori: Actually, I disagree with myself now. Now that I specify the return type with https://github.com/comit-network/comit-js-sdk/pull/176/commits/be737e03bc8df0f8f59909dbb8cf499e269ad736 (which we definitely should do) it becomes annoying to later on compose the promises in a similar way.

Opting for awaiting at the end as you originally suggested, and returning `Promise<void>` instead.

PR #179, opened 2020-03-31
Replace timeout function with external dependency

### LINKED ISSUE #120: Test `tryExecuteSirenAction`
## Problem

Currently, the logic in the class `Swap` is not tested as much as we want. That means, if an error occurs somewhere in this logic, in the "best" case the client just crashes, in the worst case we would do something which could harm the user (e.g. accidentally overfund, or underfund, etc )

## Goal

Be robust and maintain it by reviewing how to make the sdk robust to runtime failures.

## Recommendation

Test that `tryExecuteSirenAction` actually times out.

Note: `doLedgerAction` is used in e2e test and has very little logic so there is no use testing it more.
Same for `fieldValueResolver`.

luckysori:
Fixes #120 (imo).

## Straight from the commit

Motivated by the combination of wanting to test `tryExecuteSirenAction` and realising that we would just be testing `timeoutPromise`. We have already decided to replace exactly the same function in
`comit-rs` (https://github.com/comit-network/comit-rs/pull/2330#discussion_r399873316), so replacing it here has the nice consequence of not needing to write a new test, since it's not our own code.

Side-effects:

- Move `TryParams` to `swap.ts` since it's directly related and only used for executing a swap.
- Rename `util/timeout_promise.ts` to `util/sleep.ts`, since it now only holds a `sleep` function.

## Notes

I don't believe this requires a changelog update. Am I wrong?


luckysori:
> Nice work!
> Correct me if I'm wrong: Globally the `TryParams` are exported the same way?
> Because we do use them in the example. If the import in the examples has to be adapted this would be a breaking change.

Good point. I think it's exported the same, but I will quickly check locally.

luckysori:
> > Nice work!
> > Correct me if I'm wrong: Globally the `TryParams` are exported the same way?
> > Because we do use them in the example. If the import in the examples has to be adapted this would be a breaking change.
> 
> Good point. I think it's exported the same, but I will quickly check locally.

`create-comit-app` tests pass with these changes.

PR #183, opened 2020-04-13
Replace bcoin wallet with bitcoind wallet

### LINKED ISSUE #177: Bitcoin shows incorrect balances
# Problem 
When retrieving the maker & taker bitcoin balances from the SDK, it appears to not return the correct balance.

Before a 10 DAI -> 1 BTC Swap:

```
Maker: 100 DAI, 10 BTC
Taker: 100 DAI, 10 BTC
```

After a 10 DAI -> 1 BTC Swap:

```
Maker: 110 DAI, 10 BTC
Taker: 90 DAI, 10 BTC
```

# Goal

Have up2date and correct wallet balances. 

## Proposed Solution

In the `getBalance` method we should check if wallet and chain (bcoin `chain` object) are on the same hight. If not then we have to wait for syncing.

Alternatively we could return the blockhight additionally to the balance. Then the user would have to tackle it in the solution.

luckysori:
Tentatively fixes #177.

- Only after playing around with a little example inside `create-comit-app` (trying to keep the HD key functionality) did I realise that there is a `bitcoin-core` JS library that we use in `comit-rs` to instantiate a Bitcoin RPC client. I still kind of prefer to use `axios`, since we have to maintain our own types to use that other library anyway, but I may be wrong.
- These changes have been tested against a modified branch of `comit-rs`. See comit-network/comit-rs#2446.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Added
+- Implemented `BitcoinWallet` using `bitcoind`.
+
+## Deprecated
+- Implementation of `BitcoinWallet` using `bcoin`.
+
 ## [0.15.2] - 2020-03-30
 
 ### Added
--- a/src/wallet/bitcoin.ts
+++ b/src/wallet/bitcoin.ts
@@ -1,5 +1,7 @@
+import axios, { AxiosInstance, Method } from "axios";
 import { Amount, Chain, Network, Pool, TX, WalletDB } from "bcoin";
 import Logger from "blgr";
+import { toBitcoin } from "satoshi-bitcoin";
 
 /**
  * Interface defining the Bitcoin wallet functionalities needed by the SDK to execute a swap involving Bitcoin.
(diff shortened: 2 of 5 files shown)


src/wallet/bitcoin.ts
```diff
@@ -15,152 +17,153 @@ export interface BitcoinWallet {
...
-      walletdb.addTX(tx);
-    });
-
-    pool.on("block", (block: any) => {
-      walletdb.addBlock(block);
-      if (block.txs.length > 0) {
-        block.txs.forEach((tx: any) => {
-          walletdb.addTX(tx);
-        });
+    let descriptor = `wpkh(${key.toBase58()}/0h/0h/*h)`;
```
bonomat, line 78: We just need to remember that we use the same derivation path in create-comit-app or make it optionally configurable. 

See here: 
https://github.com/comit-network/create-comit-app/blob/eac523791b23d4c5cbf3a4e6af1a5f9b8b901c25/scripts/src/docker/bitcoin.rs#L143-L148

I'm pro changing it in create-comit-app to `m/0'/0'/0'` as the otherone seemed to be really weired.

src/wallet/bitcoin.ts
```diff
@@ -15,152 +17,153 @@ export interface BitcoinWallet {
...
+          jsonrpc: "1.0",
+          method: "listwallets"
+        },
+        { auth }
+      )
+      .then(res => res.data.result.includes(walletName));
+
+    if (!walletExists) {
+      await axios.post(
+        httpUrl,
```
bonomat, line 68: Oh nice, so axios is able to deal with this... way better solution then mine :)

src/wallet/bitcoin.ts
```diff
@@ -15,152 +17,153 @@ export interface BitcoinWallet {
...
-
-    for (let i = 0; i < 100; i++) {
-      pool.watchAddress(await account.deriveReceive(i).getAddress());
-      pool.watchAddress(await account.deriveChange(i).getAddress());
+    httpUrl: string,
+    username: string,
+    password: string,
+    network: Network,
+    hdKey?: string,
+    rescan?: boolean
```
bonomat, line 45: We should probably make an object out of this.

src/wallet/bitcoin.ts
```diff
@@ -15,152 +17,153 @@ export interface BitcoinWallet {
...
+    };
+
+    await axios.request({
+      ...rpcRequestConfig,
+      data: {
+        jsonrpc: "1.0",
+        method: "importmulti",
+        params: [
+          [{ desc: descriptor, timestamp: 0, range: 0 }],
+          { rescan: !!rescan ? rescan : false }
```
bonomat, line 105: instead of having this complicated logic here, I think you could just default to false in the parameter :)

src/wallet/bitcoin.ts
```diff
@@ -169,27 +172,28 @@ export class InMemoryBitcoinWallet implements BitcoinWallet {
...
+
+    if (res.data.result.chain !== network) {
+      return Promise.reject(
+        `This wallet is only connected to the ${network} network and cannot perform actions on the ${network} network`
       );
     }
   }
 }
+
+export type Network = "bitcoin" | "testnet" | "regtest";
```
bonomat, line 199: I see this already making problems :D 
See this bug: https://github.com/comit-network/comit-rs/issues/2438

I'm not sure how we want to solve it in the long run, but since you are using `getblockchaininfo` the accepted values are: 

```suggestion
export type Network = "main" | "test" | "regtest";
```

luckysori: Oh, no! Never mind then...

src/wallet/bitcoin.ts
```diff
@@ -15,152 +17,163 @@ export interface BitcoinWallet {
...
+  hdKey?: string;
+  rescan?: boolean;
+}
+
 /**
- * Dummy Bitcoin wallet based on [bcoin]{@link https://github.com/bcoin-org/bcoin}.
+ * Instance of a bitcoind 0.19.1 wallet. It uses a HD key to generate addresses, which will be derived using
+ * the Bitcoin Core wallet default derivation path: m/0'/0'/i'.
  *
  * This is to be used for demos, examples and dev environment only. No safeguards are applied, no data is written on
```
thomaseizinger, line 44: Does this still apply?

src/wallet/bitcoin.ts
```diff
@@ -15,152 +17,163 @@ export interface BitcoinWallet {
...
+    network,
+    hdKey,
+    rescan = false
+  }: BitcoindWalletArgs): Promise<BitcoindWallet> {
+    const auth = { username, password };
+
+    const key = !!hdKey
+      ? bip32.fromBase58(hdKey)
+      : bip32.fromSeed(
+          crypto.randomBytes(32),
```
thomaseizinger, line 62: Eh, do we really need this? I think the caller should just decide on the key.

src/wallet/bitcoin.ts
thomaseizinger: The "really weird" one is BIP44 compliant though 😅

https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki

Although I didn't know about the cointype to be fair. We should probably use a different number there and not 0 or 1 because we are in regtest mode.

src/wallet/bitcoin.ts
```diff
@@ -15,152 +17,163 @@ export interface BitcoinWallet {
...
-  ) {}
+  private constructor(private rpcRequestConfig: AxiosRequestConfig) {}
 
   public async getBalance(): Promise<number> {
-    const balance = await this.wallet.getBalance();
-    // TODO: Balances stay unconfirmed, try to use bcoin.SPVNode (and set node.http to undefined) see if it catches the confirmations
-    const amount = new Amount(balance.toJSON().unconfirmed, "sat");
-    return amount.toBTC(true);
+    const res = await axios.request({
+      ...this.rpcRequestConfig,
```
thomaseizinger, line 127: You can just make an axios instance instead of doing this :)

Check what we do in the `Cnd` class.

src/wallet/bitcoin.ts
```diff
@@ -15,152 +17,163 @@ export interface BitcoinWallet {
...
+          method: "getdescriptorinfo",
+          params: [descriptor]
+        },
+        { auth }
+      )
+      .then(res => res.data.result);
+    descriptor = `${descriptor}#${descriptorInfo.checksum}`;
+
+    const rpcRequestConfig = {
+      url: `${httpUrl}/wallet/${walletName}`,
```
thomaseizinger, line 106: OMG, the way they mix JSON-RPC and REST here is disgusting :D

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Changed
+- **Breaking API Change**: Restricted type of `BitcoinWallet`'s `network` parameter to improve type safety.
+- **Breaking API Change**: Replaced `InMemoryBitcoinWallet` with `BitcoindWallet`.
```
thomaseizinger, line 11: Both of these don't have to be breaking changes.

I would consider the first one to be a bugfix and the latter can be done by just deprecating `InMemoryBitcoinWallet` first and only adding the other.

Just makes for a smoother process overall because the dependency updates can happen automatically and we can just seamlessly transition :)

You don't have to change it now but I would encourage you (@bonomat / @luckysori, donno who actually wrote/decided this) to try and do things in a non-breaking way the next time :)
At the moment we are our own users so it just makes our life easier.

luckysori: Yep, you're right. I considered doing so, but I was too keen on removing bcoin. I've reverted the changes now and deprecated bcoin instead :D

Issue #94, opened 2020-01-21
Do not ask dev to know protocol steps/leverage cnd API's


D4nte:
## Problem

1. We ask the developer to manually go through the steps (sometimes [deploy](https://github.com/comit-network/comit-js-sdk/blob/5766dcc067b81e5aa5ba3c3cdc4d27973f54052e/src/swap.ts#L24), fund, redeem). Several all us fell in the trap of forgetting to do `deploy()` for ERC20 tokens.
1. We do not leverage the fact that cnd API's provide the actual available steps.

Current dev experience:
```
const swap = await takerNegotiator.takeOrder(order);

const tryParams: TryParams = { maxTimeoutSecs: 100, tryIntervalSecs: 1 };

await swap.deploy(tryParams) // If you are selling ERC20 token

await swap.fund(tryParams)

await swap.redeem(tryParams)
```

## Goal

1. Reduce the steep learning curve of doing a swap
1. Leverage cnd's API power

## Recommendation

Sit down and decide on an ideal functionality.

da-kami:
@D4nte feel free to add this to the ticket, I think it is related:

The current approach of tell-dont-ask, i.e. telling the SDK "fund my swap" and then SDK just keeps looping until the action is available and handles it is:

* Error prone.
* Not flexible/resilient enough for real world applications.
* It teaches the wrong things (calling `.redeem` suggests that you can actually do that, whereas this is not necessarily the case as we might run into a refund scenario)

Some details as provided by Thomas on Basecamp:

> If we encourage the "check next action" approach, we can build awesome stuff for our devs where simply upgrading COMIT gives their apps new functionality without them having to do something. But that is only possible if they build their apps in a certain way (hint: that way does not include hardcoding actions :D)

D4nte:
Waiting for @yosriady to finish with PoC so we can then design the ideal interface.

yosriady:
@D4nte Actually, one thing I've already noticed is for ERC20 `await swap.deploy(tryParams)` and `await swap.fund(tryParams)`, can they not be done in a single transaction instead?

D4nte:
@yosriady Do you mean one Ethereum transaction? No it cannot be done easily because you would need to deploy the htlc and then call the erc20 contract to send tokens to the htlc just deployed.
We look into doing that in one transaction and yes it might be possible but the complexity outweigh the benefits as we'd be doing some complex hacking to calculate the htlc address during its deployment to call the erc20 contract.

[...]

PR #185, opened 2020-04-16
New Swap Response Schema

D4nte:
**Round 2**

Draft for the swap response on the new lightning route `/swaps/:id`.

Most important commit is 980b3bdfa129ce887778945e9ee9f1bb0fd4f28c.
The other commit are preparation/clean-up to make it easier.

Changes from version 1
- Leverage the Siren schema as much as possible
- Remove a number of fields that are not needed (still keep them in comments for awareness).

The `//` comments are here to help reviewers and will go away before merging.

Please note that we assume that issue #94 will be done in some form or shape and that is why we allow ourselves to include less information.

Resolves comit-network/comit-rs#2159


src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+  /**
+   * Expiry: Expiry applying to this ledger, it can be an absolute or relative expiry, depending on the protocol and ledger
+   */
+  // We could have an object that allows the specification of the type of expiry here
+  expiry: number;
+}
+
+export enum EscrowStatus {
+  None = "NONE",
+  Initialized = "INITIALIZED",
```
thomaseizinger, line 404: The distinction between `Initialized` and `Deployed` is too vague for me. I would prefer if we can put some more thought into this before committing to something here.

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+  /**
+   * Asset: "bitcoin", "ether" or { standard: "erc20", contract_address: "0x..." } for non-native tokens
+   */
+  // Not sure this is ideal, let's discuss and see what we can come up with
+  asset: string | { standard: "erc20"; contract_address: "0x..." };
+  /**
+   * Expiry: Expiry applying to this ledger, it can be an absolute or relative expiry, depending on the protocol and ledger
+   */
+  // We could have an object that allows the specification of the type of expiry here
+  expiry: number;
```
thomaseizinger, line 399: Do we need to return this? I feel like that is a technical detail that is unlikely to be used.

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+  /**
+   * Network of the ledger: "mainnet", "testnet", "regtest" for Bitcoin, a chain id number for Ethereum
+   */
+  // This is tricky, an alternative would be to have more nesting where each ledger type (Bitcoin, Ethereum, Lightning) have a specific interface to declare network and asset
+  network: string | number;
+  /**
+   * Asset: "bitcoin", "ether" or { standard: "erc20", contract_address: "0x..." } for non-native tokens
+   */
+  // Not sure this is ideal, let's discuss and see what we can come up with
+  asset: string | { standard: "erc20"; contract_address: "0x..." };
```
thomaseizinger, line 394: Instead of returning raw data, we could opt to format this for the user already. Likely, they will want to see something like "0.001 BTC" or "52530 PAY".

src/cnd/cnd.ts
thomaseizinger: Since this is only for presentational purposes, maybe we can omit some of the nifty details here and decide on a smaller, fixed set of values?

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+ */
+export interface LedgerState {
+  /**
+   * The address of the htlc.
+   */
+  htlc_location?: any;
+  /**
+   * The transaction ID of the deployment transaction.
+   */
+  deploy_tx?: string;
```
thomaseizinger, line 423: This is all going to be `null` for `halight` right?

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+export interface FullSwapProperties {
+  /**
+   * The local id of the swap.
+   */
+  local_id: string;
+  /**
+   * The parameters of this swap.
+   */
+  // Decided to put everything together as now the protocols are tightly linked to the ledgers and assets
+  parameters: {
```
thomaseizinger, line 455: I would be in favor of removing as much nesting as possible.

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+    beta: Parameters;
+  };
+  /**
+   * The status this swap is currently in.
+   */
+  status: SwapStatus;
+  /**
+   * The data learned at creation of the swap.
+   */
+  created: {
```
thomaseizinger, line 466: I would remove this nesting, I don't think a client would care about this.

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+    /**
+     * The role in which you are participating in this swap.
+     */
+    role: "Alice" | "Bob";
+    /**
+     * The peer-id and hint address of the counterparty of this swap.
+     */
+    peer: Peer;
+    self_identities: {
```
thomaseizinger, line 475: Do we have to tell the client what they told us?

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+  // While it could seem odd to have created/finalized section, it does simplify by reducing the number of conditional properties
+  finalized?: {
+    /**
+     * The shared id of the swap.
+     */
+    shared_id: string;
+    /**
+     * The secret hash
+     */
+    secret_hash: string;
```
thomaseizinger, line 506: What does the client do with technical information like this?

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+   *
+   * This section is only available once a swap is marked as "finalized", at the end of the communication phase.
+   * Which means it is not present when the status is "CREATED"
+   */
+  // While it could seem odd to have created/finalized section, it does simplify by reducing the number of conditional properties
+  finalized?: {
+    /**
+     * The shared id of the swap.
+     */
+    shared_id: string;
```
thomaseizinger, line 502: Do we have to share this one?

src/cnd/cnd.ts
thomaseizinger: Can anyone think of an application on top of COMIT that would need structured data here _after_ the swap has already started?
It doesn't have to be a string btw, we could go for something like:

`{
	"quantity": "52530",
	"symbol": "PAY"
}`

D4nte:
> Good effort on picking this one up :)

I really did not want to be the one designing all endpoints but hey, it's life.

> My feedback can be summarized as: Unless we have a very strong reason to return some of the information, I would omit it for now.

Cool, will action that

> Maybe we can structure this state into more levels, i.e. different resources that the client can explore if they really need to but generally I think we are better off for now to keep the API to a minimum.

Hum, in your other comment you mention less nesting. What kind of levels are you talking about? I can only think of nesting levels. When you say resources, are you talking of endpoints?

> Probably worth having a chat with @yosriady on which data he is currently using and how.

Yep. He is tagged to review and I mentioned it was coming already.

thomaseizinger:
> Hum, in your other comment you mention less nesting. What kind of levels are you talking about? I can only think of nesting levels. When you say resources, are you talking of endpoints?

Yes! Just a vague idea but maybe we can do something like:

GET /swaps/:id

- returns data common to all swaps
- contains links to:
	- blockexplorer for individual transactions? 
	- a link to a resource describing the alpha-ledger-state (this could again be a siren-document that has certain `class` attributes like `["han-ethereum", "state"]` so we can have the properties fixed

Thinking about it, we could actually also use siren's [sub-entities](https://github.com/kevinswiber/siren#entities-1) for this!

Note that the two approaches are actually compatible: Siren differentiates between linked sub-entities and embedded ones so the states can be addressed with dedicated URLs if we want to.

For now, I wouldn't do any of this though but just strip down the response to a bare minimum and try to find a model that works for all ledgers/assets.

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+   */
+  protocol: string;
+  /**
+   * Ledger: "ethereum", "bitcoin" or "lightning"
+   */
+  ledger: string;
+  /**
+   * Network of the ledger: "mainnet", "testnet", "regtest" for Bitcoin, a chain id number for Ethereum
+   */
+  // This is tricky, an alternative would be to have more nesting where each ledger type (Bitcoin, Ethereum, Lightning) have a specific interface to declare network and asset
```
thomaseizinger, line 388: Bitcoin's "magic bytes" can be interpreted in a u64 so they could also be a number if we want. Alternatively, we could use the name of the Ethereum network (Ropsten, etc).

This sort of brings back the discussion of how to identify networks. We are not using magic-bytes for Bitcoin at the moment for example ...

src/cnd/cnd.ts
yosriady: I think we will need this to allow apps to handle letting a swap expire.

src/cnd/cnd.ts
thomaseizinger: We don't have to maintain a mapping for this (should likely cache it though). We can ask the ERC20 contract for it's symbol via `eth_call`.

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+ * Element of the array in the payload returned when fetching all swaps on the `/swaps/` endpoint
+ */
+export interface SwapElementResponse extends EmbeddedRepresentationSubEntity {
+  properties?: PartialSwapProperties;
+}
+
+/**
+ * The overall status of a swap
+ */
+export enum SwapStatus {
```
yosriady, line 349: One issue I had with the current `Swap` status fields, which is similar to this is that it was too high level. 

I had to parse and process `communication`, `alpha_ledger`, and `beta_ledger` properties to really know 'where' the swap is currently at. What do you think of making these more granular (e.g. `MAKER_FUNDED`, `TAKER_DEPLOYED`, etc.).

thomaseizinger:
> When you say resources, are you talking of endpoints?

Basically yes. In REST, a resource is the concept that is _addressed_ through a URL. A resource can be in different _states_ over time and when you do a GET request on a resource it gives you a _representation_ of it at that current point in time.

Hence the name: **RE**presentational **S**tate **T**ransfer.

When you do a POST/PUT for example, you are _transferring_ a _representation_ of a state for a given resource with the semantic difference that POST says: Create a new resource vs PUT says modify the existing resource I am addressing through the URL.

src/cnd/cnd.ts
yosriady: I think contract address for ERC20 tokens is required. The token metadata itself is optional since you can query it yourself, but having it included is helpful.

src/cnd/cnd.ts
yosriady: I second the above point.

src/cnd/cnd.ts
yosriady: Ah ok, it's `EscrowStatus`?
Could we consolidate the two statuses?

thomaseizinger: The thing is, `cnd` cannot really know who is the taker and who is the maker in a swap unless you tell it. We only know about cryptographic roles and ledgers but not about who made the initial offer :)

That being said, we can still enrich this status to say more about the individual swap ledgers although the original idea was to keep this really high-level.

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+  // Not sure this is ideal, let's discuss and see what we can come up with
+  asset: string | { standard: "erc20"; contract_address: "0x..." };
+  /**
+   * Expiry: Expiry applying to this ledger, it can be an absolute or relative expiry, depending on the protocol and ledger
+   */
+  // We could have an object that allows the specification of the type of expiry here
+  expiry: number;
+}
+
+export enum EscrowStatus {
```
yosriady, line 402: I think this will be useful, at least for the SDK's state machine to know what to do next.

src/cnd/cnd.ts
yosriady: What does 'created' mean?

src/cnd/cnd.ts
```diff
@@ -329,6 +329,218 @@ export interface SwapProperties {
...
+    };
+  };
+  /**
+   * The data learned during the communication phase.
+   *
+   * This section is only available once a swap is marked as "finalized", at the end of the communication phase.
+   * Which means it is not present when the status is "CREATED"
+   */
+  // While it could seem odd to have created/finalized section, it does simplify by reducing the number of conditional properties
+  finalized?: {
```
yosriady, line 498: I'm not sure I understand the point of splitting `created` and `finalized?`.

src/cnd/cnd.ts
thomaseizinger: Well, clients should never make assumptions about what to do next based on the state :grimacing: 

They should only act on the precense/absence of actions otherwise you are re-implementing the state machine we already have in the backend.

These states should IMO be used purely for presentational purposes :)

src/cnd/cnd.ts
yosriady: I think including both the `self` and `counterparty` will be useful for monitoring / dashboards. Also, the client could have switched wallets in between.

src/cnd/cnd.ts
D4nte: > I think we will need this to allow apps to handle letting a swap expire.

Actually this should be done via the refund action, also see the pitched idea to not have the dev know what are the steps of a swap.

src/cnd/cnd.ts
D4nte: Well yes but this design intends to support han, herc20 and halight.

src/cnd/cnd.ts
D4nte: I'll remove.

src/cnd/cnd.ts
D4nte: I actually decided to remove it, the network is not something dynamic really.

src/cnd/cnd.ts
D4nte: Hum, not really sure about that. Is it really worth duplicating the information in the response?

src/cnd/cnd.ts
D4nte: Yes the idea is for presentational purposes but having a static list allows you the app to show a specific translation/icon/explanation for each status to the user.

src/cnd/cnd.ts
D4nte: Indeed, removed.

src/cnd/cnd.ts
D4nte: I think this set of value makes sense because of another train of thought I had recently. I have added documentation to clarify those.

src/cnd/cnd.ts
D4nte: At this stage I am re-using the existing `Asset` class as it contains the contract. I am not convinced cnd should be fetching such information from the blockchain. I think it makes more sense for cnd to provide the raw information upon which the protocol is executed.

src/cnd/cnd.ts
D4nte: I actually removed them when working on version 2, I have no re-added them as per Yos' request.

src/cnd/cnd.ts
thomaseizinger: Well it is always a trade-off right. The reason I was suggesting this is because it simplifies the API, which is what we are trying to do here.

If we say, the contract address is required, then my question would be required for what?

I am fine if we can name usecases that will need the address for things that we cannot or don't want to provide as cnd.

> I am not convinced cnd should be fetching such information from the blockchain.

We can include a static list if we don't want to fetch the information. But the source is not what I would want to discuss. What I want to discuss is the reason we are including the information.

Any conditional logic that we can avoid for the client (`if asset.kind == 'erc20' { ... }`) is a win in terms of how easy the API is to use. Every polymorphic field in the response is going to cause some kind of conditional in the client ..

src/cnd/cnd.ts
thomaseizinger: When are we building monitoring or dashboards?

I don't think it is on the timeline anytime soon and even if it were/will be, then I would like to see a concrete usecase that requires `cnd` to share this information with the client.

I would assume that any wallet used by the user is tracking money by themselves and we don't need to tell it the addresses again.

thomaseizinger: We can always add fields to this response later but once they are there, we need to maintain them and removing them is a backwards-incompatible change.

Just trying to make our own life easier :)

src/cnd/cnd.ts
D4nte: @yosriady what's your input on this topic please? I consider it unresolved at this stage :)

src/cnd/swaps_payload.ts
```diff
@@ -79,3 +80,227 @@ export type HalightLightningBitcoinHerc20EthereumErc20RequestBody = CoreRequestB
...
+}
+
+/**
+ * The identities related to a given party.
+ */
+export interface Identities extends EmbeddedRepresentationSubEntity {
+  /**
+   * The relation of these identities to the parent object (*SwapProperties).
+   */
+  rel: ["self" | "peer"];
```
thomaseizinger, line 283: self is actually a registered relation and means something very different :)

src/cnd/swaps_payload.ts
```diff
@@ -79,3 +80,227 @@ export type HalightLightningBitcoinHerc20EthereumErc20RequestBody = CoreRequestB
...
+   * To protect the user, if an incorrect amount of asset have been sent to the escrow, cnd will not propose redemption
+   * as an option and only the refund actions will be available down the line.
+   */
+  IncorrectlyFunded = "INCORRECTLY_FUNDED"
+}
+
+/**
+ * The identities related to a given party.
+ */
+export interface Identities extends EmbeddedRepresentationSubEntity {
```
thomaseizinger, line 279: I don't think we should return this, these are technical details of the swap that I am not sure what the client should do with them.

src/cnd/swaps_payload.ts
```diff
@@ -79,3 +80,227 @@ export type HalightLightningBitcoinHerc20EthereumErc20RequestBody = CoreRequestB
...
+  properties: {
+    /**
+     * Protocol: "han", "herc20" or "halight"
+     */
+    protocol: string;
+    /**
+     * Ledger: "ethereum", "bitcoin" or "lightning"
+     */
+    ledger: string;
+    // network: string | number; // No network as an app should only handle one type of network at a given time
```
thomaseizinger, line 176: Interesting idea :)

src/cnd/swaps_payload.ts
```diff
@@ -79,3 +80,227 @@ export type HalightLightningBitcoinHerc20EthereumErc20RequestBody = CoreRequestB
...
+  title: "Parameters of the Alpha Ledger" | "Parameters of the Beta Ledger";
+  /**
+   * Class of this sub-entity to facilitate parsing.
+   */
+  class: ["parameters"];
+  properties: {
+    /**
+     * Protocol: "han", "herc20" or "halight"
+     */
+    protocol: string;
```
thomaseizinger, line 171: You could make the protocol a class, then the properties are fixed.

thomaseizinger: You could also merge the `LedgerState` into this then because we can guarantee a set of fixed fields per protocol, i.e. `transactions.deploy` for example.

thomaseizinger: The protocols would have to be:

- `han-ethereum`
- `han-bitcoin`
- `herc20`
- `halight`

but that is how I like to think about them anyway :)

Also gets rid of the `ledger` field.

thomaseizinger: You could then expand the `asset` field into:

- `quantity`
- `contract_address` (only for `herc20`)

src/cnd/swaps_payload.ts
D4nte: What do you think would make more sense?
- local
- own
- ..?

src/cnd/swaps_payload.ts
D4nte: what happens when we support halight for litecoin?

D4nte: Is that the proposal you are making?

```
{
  class: ["parameters","han-ethereum"|"han-bitcoin"|...];
  properties: NativeAsset|Erc20Asset;
...
}

export interface NativeAsset {
	name: "bitcoin"|"ether";
	quantity: string;
}

export interface Erc20Asset {
	contract_address: "0x...";
	quantity: string;
}
```

thomaseizinger: Yes with the change that I wouldn't do `NativeAsset` I think :)

The way I like to think about `han` is that it is an abstract protocol that can be instantiated for ethereum and bitcoin but it cannot appear by itself, hence I would always talk about `han-ethereum` or `han-bitcoin`.

D4nte: > Yes with the change that I wouldn't do `NativeAsset` I think :)
> 
> The way I like to think about `han` is that it is an abstract protocol that can be instantiated for ethereum and bitcoin but it cannot appear by itself, hence I would always talk about `han-ethereum` or `han-bitcoin`.

Can you please write some code, I don't understand what you mean. How do you express bitcoin asset if you dont do `NativeAsset`?

D4nte: @yosriady @thomaseizinger can you please both clarify your position on this subject?

src/cnd/swaps_payload.ts
D4nte: identities removed

src/cnd/swaps_payload.ts
D4nte: Removed.

src/cnd/swaps_payload.ts
thomaseizinger: You only need the quantity field if you have `han-bitcoin` :)

thomaseizinger: > > The protocols would have to be:
> > 
> >     * `han-ethereum`
> > 
> >     * `han-bitcoin`
> > 
> >     * `herc20`
> > 
> >     * `halight`
> > 
> > 
> > but that is how I like to think about them anyway :)
> > 
> > Also gets rid of the `ledger` field.
> 
> what happens when we support halight for litecoin?
> 

Name it `halight-bitcoin` then if you want :)

src/cnd/swaps_payload.ts
```diff
@@ -0,0 +1,312 @@
...
+  rel: ["alpha" | "beta"];
+  /**
+   * Human readable title.
+   */
+  title: "Parameters of the Alpha Ledger" | "Parameters of the Beta Ledger";
+  /**
+   * Class of this sub-entity to facilitate parsing.
+   */
+  class: ["parameters"];
+  properties: HanEthereum | HanBitcoin | Herc20 | HalightBitcoin;
```
D4nte, line 169: @yosriady What do you think about that? Would it be a nightmare to handle?

yosriady: I think this is fine.

D4nte:
It looks like this is ready to be merged, I'll leave it until Wed 12:00 to squash commits and merge.

PR #187, opened 2020-04-19
Add label-slow-prs GitHub Action

luckysori:


luckysori:
> How can I see the action `coblox/label-slow-prs@v0.1.0` ?

The repo is here: https://github.com/coblox/label-slow-prs. The action only runs once a week and it should be triggered at around noon, so you will be able to see the workflow in action then :crossed_fingers:

PR #194, opened 2020-04-23
Trigger checks once release PR is created

D4nte:
GitHub Action does not allow recursion so the
CI checks are not triggers on the release branch
as it is pushed by the "draft new release" action.
So we use repo dispatch to make it happen.


D4nte:
> That actually works? You are a mad man Franck!

We shall see!

PR #225, opened 2020-04-28
Assert chain_id on actions against Ethereum

### LINKED ISSUE #4: Caller cannot specify a network for deployContract and callContract on the EthereumWallet
## Problem

A caller of `deployContract` or `callContract` (both functions on the EthereumWallet) currently **cannot** specify the network (mainnet, ropsten, etc) on which the action should be performed.

https://github.com/comit-network/comit-js-sdk/blob/e8d0fc499af3614cb350776ebc60a4494e38bb64/src/ethereumWallet.ts#L23-L39

This can be dangerous because funds could accidentally be sent on the wrong network, i.e.: the wallet is connected to a Mainnet node but the application would actually like to interact with a ropsten node

## Goal

Make sure the network is not implicit so that the application has to actively specify it.

## Recommendation

We should add a `chainId` parameter to these methods and assert that the underlying `Web3` client is actually connected to a network with that chainId.

bonomat:
cnd tells us against which network a particular action should be executed.
This is now enforced by the sdk.

Resolves #4 

Does this count as a breaking change? The functions are public but we only use it internally.

### DIFF
--- a/src/wallet/ethereum.ts
+++ b/src/wallet/ethereum.ts
@@ -119,7 +119,7 @@ export class EthereumWallet {
 
     if (actualNetwork.chainId !== expectedChainId) {
       return Promise.reject(
-        `This wallet is connected to the chainId with chainId: ${expectedChainId}  and cannot perform actions on against a chainId with ${actualNetwork.chainId} chainId`
+        `This wallet is connected to the chain with chainId: ${expectedChainId}  and cannot perform actions on chain with chainId ${actualNetwork.chainId}`
       );
     }
   }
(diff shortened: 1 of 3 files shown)


src/wallet/ethereum.ts
```diff
@@ -109,4 +113,14 @@ export class EthereumWallet {
 
     return response.hash;
   }
+
+  private async assertNetwork(expectedChainId: number): Promise<void> {
+    const actualNetwork = await this.wallet.provider.getNetwork();
+
+    if (actualNetwork.chainId !== expectedChainId) {
+      return Promise.reject(
+        `This wallet is connected to the chainId with chainId: ${expectedChainId}  and cannot perform actions on against a chainId with ${actualNetwork.chainId} chainId`
```
da-kami, line 122:
```suggestion
        `This wallet is connected to the chain with chainId: ${expectedChainId}  and cannot perform actions on chain with ${actualNetwork.chainId} chainId`
```

PR #229, opened 2020-04-29
`sendPayment` returns when payment is `in flight`

### LINKED ISSUE #180: Resolve `fund` promise when paying lightning hold invoice
## Problem

Currently, the promise returned by `LightningWallet.sendPayment()` setups the payment over lightning but does not resolve until the payment is settled. The settlement happens at a later stage when the other party reveal the secret.

https://github.com/comit-network/comit-js-sdk/blob/5b893fd723940776aab01c71ca298a6308ea2895/src/wallet/lightning.ts#L31

The issue is that when using this function, we do not know when the payment is setup (or "in flight" in lightning terms). We have to park the promise and wait "some time" to proceed forward. Which is not great and manifest itself with the use of sleep in the comit-rs e2e tests:

https://github.com/comit-network/comit-rs/blob/ed0857d19beba08ce059d04c898c1c62bb5ecbbf/api_tests/tests/ether_halight.ts#L18

## Goal

Modify the `sendPayment()` function so that is promise resolves once the payment is "in flight" but not yet settled.

It should return a `Transaction` (extend the one from `transaction.ts`) object that should then allow the lib consumer to wait until the payment is settled.

## Recommendation

We currently use `sendPaymentSync` from ln-rpc. Instead, we should probably use `sendPayment` and see if we can get a relevant status from it.
If not, then poll on `listPayments` and resolve the Promise once the Payment is marked `IN_FLIGHT`.

D4nte:
The promise returned by `LightningWallet.sendPayment`
now resolves when the payment is `in flight`,
previously resolving when `settled`.

The promise resolves to an async function that itself,
returns when the payment is `settled`.

See https://github.com/comit-network/comit-rs/pull/2553 on how to use the API.

Resolves #180

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
--  Add new function (`getAllSwaps`) to `ComitClient` which provides the means of retrieving all swaps regardless of their state. 
+-  Add new function (`getAllSwaps`) to `ComitClient` which provides the means of retrieving all swaps regardless of their state.
+
+### Changed
+
+-   **Breaking API Change**: The promise returned by `LightningWallet.sendPayment` now resolves when the payment is `in flight`, previously resolving when `settled`. 
 
 ## [0.15.6] - 2020-04-27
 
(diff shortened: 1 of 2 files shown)


D4nte:
> Interesting solution!

Got inspired by the lock file lib 😁

PR #231, opened 2020-05-01
Improve error messages generated from problems

thomaseizinger:
While doing https://github.com/comit-network/comit-rs/pull/2554, I realized that the error messages generated by the problems were actually not that great because I was just using the title of the problem.

This patch aims to make the situation better and also adds tests for the messages that are generated.

### DIFF
--- a/src/cnd/axios_rfc7807_middleware.spec.ts
+++ b/src/cnd/axios_rfc7807_middleware.spec.ts
@@ -1,0 +1,49 @@
+import { Problem } from "./axios_rfc7807_middleware";
+
+describe("problem message", () => {
+  it("should format all fields to a descriptive message", () => {
+    const problem = new Problem({
+      title: "Swap not supported.",
+      status: 400,
+      detail: "This combination of protocols is not supported.",
+      type: "https://comit.network/docs/errors/swap-not-supported"
+    });
+
+    expect(problem.message).toMatchInlineSnapshot(
+      `"Request failed with status code 400: Swap not supported. This combination of protocols is not supported. See https://comit.network/docs/errors/swap-not-supported for more information."`
+    );
+  });
+
+  it("should exclude the url if it is not present", () => {
+    const problem = new Problem({
+      title: "Swap not supported.",
+      status: 400,
+      detail: "This combination of protocols is not supported."
+    });
+
+    expect(problem.message).toMatchInlineSnapshot(
+      `"Request failed with status code 400: Swap not supported. This combination of protocols is not supported."`
+    );
+  });
+
+  it("should exclude the statuscode if it is not present", () => {
+    const problem = new Problem({
+      title: "Swap not supported.",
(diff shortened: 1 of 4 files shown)


src/cnd/axios_rfc7807_middleware.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { Problem } from "./axios_rfc7807_middleware";
+
+describe("problem message", () => {
+  it("should format all fields to a descriptive message", () => {
+    const problem = new Problem({
+      title: "Swap not supported.",
+      status: 400,
+      detail: "This combination of protocols is not supported.",
+      type: "https://comit.network/docs/errors/swap-not-supported"
```
thomaseizinger, line 9: One day, I hope that we can document all those errors as part of our docs and fill those URLs into the error responses of cnd to present users with beautiful, helpful messages <3

da-kami, line 9: This page does not exist, are we planning to add it?
Should this be a link to the sdk documentation?

da-kami: I would like that too, but for now I would only add links that point to something that exists already.
If a dev receives such an error and tries to use the URL it will cause more confusion than it helps at the moment.

da-kami: Would it always exclude that at the moment because the URL is not present?

src/cnd/axios_rfc7807_middleware.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    });
+
+    expect(problem.message).toMatchInlineSnapshot(
```
D4nte, line 12: I am not sure to understand why you are using the snapshot feature here as `message` is just a string with no html elements.

src/cnd/axios_rfc7807_middleware.spec.ts
thomaseizinger: > Would it always exclude that at the moment because the URL is not present?

If the URL is not present in the response, nothing in regards to the URL is part of the output.

The response comes from `cnd` so we have to start returning them there once they exist. The SDK never "assembles" these links (check the code on how this message is generated)!

thomaseizinger: This is just test data, so I wouldn't care too much.

I can replace it with something completely non-nonsensical (like `example.com`) if you want :)

src/cnd/axios_rfc7807_middleware.spec.ts
thomaseizinger: Because I was too lazy to write the expected message :)

I wrote the code first and I am using snapshot tests to "freeze" the implementation after I confirmed visually that this is what I want!

PR #240, opened 2020-05-28
Use protocol names to simplify swap creation + add hbit

da-kami:
This is prep-work to clean up the cnd e2e tests and enable `herc20-hbit`/`hbit-herc20`

* Reduce all names of `create` routes for swaps and corresponding classes to only use the protocol names.
* Remove `han`
* Remove all `Ether` related stuff (won't be supported atm)
* Add `hbit` related stuff


da-kami:
Sweet! The api-changes labelling bot works great 😁

da-kami:
> I think we initially said to iterate on this in the e2e tests, I guess this didn't happen :)

Well, we did, but I decided to go like this because I think it is faster. Otherwise we have to decide what to bring in from the SDK into the e2e tests so we can communicate with cnd. 

OK if we roll with this, or would you insist on doing it in the e2e test? 
I feel moving this into the e2e tests it not really worth the effort.

PR #241, opened 2020-06-02
Recommended Unique Action

D4nte:
Update the SDK to support cnd's recommended unique action as per the project.

See the usage in e2e: https://github.com/comit-network/comit-rs/pull/2765

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 -   Support for `hbit-herc20` and `herc20-hbit` swaps.
+-   New `Swap.nextAction` interface, allows to get and execute the next recommended action returned by cnd 0.8.0 and above. 
 
 ### Changed
 
--- a/src/action.spec.ts
+++ b/src/action.spec.ts
@@ -1,0 +1,41 @@
+import { Action } from "./action";
+import { Step, SwapAction } from "./cnd/swaps_payload";
+import { Swap } from "./swap";
+
+const swapAction: SwapAction = { name: Step.Redeem, href: "/redeem" };
+
+describe("Action", () => {
+  it("has the same name than the swap action used to construct", () => {
+    const swap = (undefined as unknown) as Swap;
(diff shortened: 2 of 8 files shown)


D4nte:
The API changes are backward compatible.

src/action.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (!recommendedAction || this.name !== recommendedAction.name) {
+      throw new Error(
+        `Action ${
+          this.name
+        } is not recommended anymore. Current recommended action is: ${
+          recommendedAction ? recommendedAction.name : recommendedAction
+        }`
+      );
+    }
+
```
thomaseizinger, line 32: I don't think we need this check. `cnd` MUST fail anyway if you try to do an action that no longer works.

You don't get any guarantees out of this because theoretically, the action you are holding on to could get stale between your check above here and where you execute it below :)

src/action.spec.ts
```diff
@@ -0,0 +1,91 @@
+import { Action } from "./action";
+import { Step, SwapAction } from "./cnd/swaps_payload";
+import { Swap } from "./swap";
+
+const swapAction: SwapAction = { name: Step.Redeem, href: "/redeem" };
+
+describe("Action", () => {
```
thomaseizinger, line 7: I appreciate the effort for writing tests but I am not sure how much risk they are eliminating, esp. given their verbosity and mocking efforts.

With the suggestion below of letting `cnd` handle the "action is no longer recommended" scenario, I don't think we need these tests.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 -   Support for `hbit-herc20` and `herc20-hbit` swaps.
+-   New `Swap.nextAction` interface, allows to manipulate the next recommended action returned by cnd 0.8.0 and above.
```
thomaseizinger, line 13: Why the wording "manipulate"?

src/action.ts
D4nte: Ah yes indeed.

PR #242, opened 2020-06-02
Release 0.16.1

D4nte:
Trying to finish the "Recommended Unique Action" Project by resolving https://github.com/comit-network/comit-rs/issues/2753

First attempt was to simply release the current dev branch and integrate the latest release in the e2e tests but it includes [breaking changes](https://github.com/comit-network/comit-js-sdk/pull/240) I am not familiar with.

Second attempt was to bring https://github.com/comit-network/comit-js-sdk/pull/241 changes in the comit-rs e2e test. Unfortunately, this is not possible because `Swap.fieldValueResolver()` is private and the only method available to execute an action is `Swap.executeSirenAction()` which take the name of the action and will retrieve the full content from cnd again.

This is the third attempt, cherry pick the RUA changes and releasing them.

### DIFF
--- a/src/action.spec.ts
+++ b/src/action.spec.ts
@@ -1,0 +1,41 @@
+import { Action } from "./action";
+import { Step, SwapAction } from "./cnd/swaps_payload";
+import { Swap } from "./swap";
+
+const swapAction: SwapAction = { name: Step.Redeem, href: "/redeem" };
+
+describe("Action", () => {
+  it("has the same name than the swap action used to construct", () => {
+    const swap = (undefined as unknown) as Swap;
+    const action = new Action(swapAction, swap);
+
+    expect(action.name).toEqual(swapAction.name);
+  });
+
+  it("calls Swap.executeAction with the parameters of the swap action used to construct", async () => {
+    const mockExecuteAction = jest.fn(async () => {
+      return {};
+    });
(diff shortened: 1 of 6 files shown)


src/action.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      return swapAction;
+    });
+    const mockDoLedgerAction = jest.fn(async () => {
+      return "";
+    });
+    const swap = ({
+      executeAction: mockExecuteAction,
+      nextAction: mockNextAction,
+      doLedgerAction: mockDoLedgerAction
+    } as unknown) as Swap;
```
thomaseizinger, line 29: It took me a while to see how you are controlling the test input (`swapAction`) here and how the setup is related to your assertions.

This test would be easier to read if the "arrange" part is only one or two statements.

Personally, I usually don't bother writing tests that need such a complicated setup because I am too lazy :sweat_smile: 
If the underlying code is still crucial and needs testing, there is probably a way of refactoring that logic out so it is less cumbersome to access.

D4nte: I'll try to do a follow-up PR.

thomaseizinger: Not a biggy :)

Finishing RUA is more important I'd say.

PR #245, opened 2020-06-02
Export siren as siren to avoid unsuspected name clashes and override

D4nte:
I actually ended up overriding a the `Action` siren type without noticing due to the wildcard export.


D4nte:
> Does that mean the previous change was in fact a breaking change? :)

Don't tell anyone.

PR #255, opened 2020-07-29
Wait for next message from lnd if channel is not ready

D4nte:
Sometimes, lnd returns a `OpenStatusUpdate` that does not contain the outpoint of the channel.

Instead of failing, we wait for the next message.

Hopefully we don't wait forever.

Used in https://github.com/comit-network/comit-rs/pull/3020


D4nte:
It works locally with comit-rs. Integrating and releasing to test in the CI.

